PR #260, opened 2015-02-22
Fix file path case problem

haya14busa:
### Problem:
  If you use Ubuntu VM(case sensitive file system) on Mac OS(case
  in-sensitive file system) and the &runtimepath is in a shared folder,
  the case handling for paths could different with different paths.
  It could cause exception, `vital: module not found:`.

ref: https://github.com/Lokaltog/vim-easymotion/issues/171

### Solution:
  Do not use tolower() (do not call s:_unify_path()) to get scripts'
  self file path.

http://lingr.com/room/vim/archives/2015/02/22#message-21320401

### DIFF
--- a/autoload/vital/__latest__.vim
+++ b/autoload/vital/__latest__.vim
@@ -1,4 +1,5 @@
 let s:self_version = expand('<sfile>:t:r')
+let s:self_file = expand('<sfile>')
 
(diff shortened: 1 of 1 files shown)


thinca:
Thanks!

haya14busa:
:muscle: :bird:

PR #264, opened 2015-03-14
Web.HTTP: add gzip decode to python curl requests

nicwest:
This will decode websites served with gzip. 

I came accross a website that didn't respond to accept-encoding, and all the content comes back compressed.

wget has no built in decompression, so the only option would be to pipe it into gunzip or similar. probably a bit of an edge case anyway.

here is a url for testing: http://www.nic-west.com/messages/test.txt

### DIFF
--- a/autoload/vital/__latest__/Web/HTTP.vim
+++ b/autoload/vital/__latest__/Web/HTTP.vim
@@ -278,7 +278,11 @@ function! s:clients.python.request(settings) abort
 try:
     class DummyClassForLocalScope:
         def main():
-            import vim, urllib2, socket
+            try:
+                from StringIO import StringIO
+            except ImportError:
+                from io import StringIO
+            import vim, urllib2, socket, gzip
             def vimstr(s):
                 return "'" + s.replace("\0", "\n").replace("'", "''") + "'"
 
@@ -327,7 +331,15 @@ try:
 
                 st = status(res)
                 responseHeaders = st + ''.join(res.info().headers)
-                return (responseHeaders, res.read())
+                response_body = res.read()
+
+                gzip_decompress = settings.get('gzipDecompress', False)
+                if gzip_decompress:
+                    buf = StringIO(response_body)
+                    f = gzip.GzipFile(fileobj=buf)
+                    response_body = f.read()[:-1]
+
+                return (responseHeaders, response_body)
 
             (header, body) = access()
             vim.command('let header = ' + vimstr(header))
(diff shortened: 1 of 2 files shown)


thinca:
LGTM :+1:

mattn:
SGTM but it should be possible to disable gzip by `a:settings`.

nicwest:
@mattn forgive me I'm not super familar with this project yet, would something like this work better?

Should I add options to `HTTP.get()` and `HTTP.post()`?

mattn:
> Should I add options to HTTP.get() and HTTP.post()?

I'm not sure. But I'm thinking, for non-intelligent server (ex: hand-made http server that does not support compressed transport), we should better to provide a way to disable it. So it should be easy to do.

rhysd:
Hi,

Thank you for nice pull request, @nicwest !

> Should I add options to HTTP.get() and HTTP.post()?

I think there is no need to add an extra parameter to `HTTP.get()` and `HTTP.post()`.  Is it sufficient to use `request(settings)` to enable gzip decompression?  I'm worried about making parameters of them complicated.

nicwest:
@rhysd I was having similar thoughts. I think sites which will only return gzip'd data and don't respond to accept-content headers will be fairly rare, so this is a pretty edge case to start with.

rhysd:
@nicwest 

So current implementation is enough to merge, I think.

@mattn How about you?

mattn:
@rhysd do you mean `current implementation` include https://github.com/nicwest/vital.vim/commit/9529d350bac61c1ec8509a2ded6d8a79972cc01c ?

ujihisa:
Thanks! vital.vim developers don't always know very much about python, so this contribution is helpful a lot for us.
I believe you can already merge this.

rhysd:
> do you mean current implementation include nicwest@9529d35 ?

Yes.

mattn:
LGTM

kamichidu:
Great work, thanks a lot!

Commit messages has incorrect format and ''wip'' word.
@nicwest  Tell us when you are ready to merge it.

rhysd:
@kamichidu 

Thanks for pointing out.

@nicwest 

Could you please reword the latest commit message? The format is (as you know) `Web.HTTP: {message}`

nicwest:
All done cheers guys.

@ujihisa I'm a python developer, if you ever need a hand feel free to shout at me.

rhysd:
@nicwest 

Thank you :dog: :tada:

PR #305, opened 2015-06-30
RFC:WIP: System.Logging: Add a new logging module

### LINKED ISSUE #222: Let's logger
7.4.503 で writefile に "a" が使える様になったので、ロガークラスが作れるかと思います。

って事で、誰かやって！

lambdalisue:
The module is strongly inspired by [Python's logging module](http://docs.python.jp/2/library/logging.html).
The unit tests were performed with Vim 7.3 and 7.4.712

#### ToDo

- [x] Add Vim 7.3 compatible `writefile` feature to `System.File` to support Vim 7.3
- [x] Add `System.Logging` module
- [x] Add `test/System/Logging.vimspec`
- [ ] Add `doc/vital-system-logging.txt`
- [ ] Add an entry to `README.md` and `doc/vital.txt`
- [ ] Support Windows (need help. mattn? lol)
- [ ] Rebase commits to clean up the commit messages
- [ ] Write `Changes` (is it required to write for a new module?)

#### Usage (Concept)

```vim
let s:L = s:V.import('System.Logging')
let s:logger = s:L.of(expand('<sfile>'))

call s:logger.debug('%s! It should not be shown.', 'Hello world')
call s:logger.warning('%s! It should be shown.', 'Hello world')
call s:logger.exception('%s! It should be shown.', 'Hello world')

let s:root_logger = s:L.of()
call s:root_logger.error('This is a root')
```

Assume the code above in `~/vim-gita/autoload/gita/test/log.vim` and the vital module is in `~/vim-gita/autoload/vital/_vim_gita`,
then the followings will be written into `~/.vim_gita/logger.log` (for non XDG environment).

```
2015-07-01 01:22:53;gita/test/log;WARNING; Hello world! It should be shown.
2015-07-01 01:22:53;gita/test/log;ERROR; Hello world! It should be shown.
2015-07-01 01:22:53;gita/test/log;ERROR; Exception: "" in 
2015-07-01 01:22:53;.;ERROR; This is a root
```

Note that `logger.exception` use `v:exception` and `v:throwpoint` thus the part of those variables will not be shown for non exception block.

This PR will close #222


lambdalisue:
What do you think about if I add `doautocmd User vital-system-logging-log` which is useful to execute `checktime`?

ujihisa:
something like `:execute 'doautocmd User vita-system-logging-log-' . vital_version` would be handy for future compatibility

rhysd:
@lambdalisue 

Nice module! Thank you for development.

Why you use `of`? Python's `Logging` module doesn't seem to use the name.  And `Logging.of({name})` seems not to make sense (In addition, it can omit the argument, `Logging.of()` ).  I think it should be renamed.

> Write Changes (is it required to write for a new module?)

No need to write Changes for new modules.  And please do not forget to add this module to README.

rhysd:
> What do you think about if I add doautocmd User vital-system-logging-log every after a new record is written into a logfile. It would be useful to execute checktime I think.

Vital is embedded in a plugin.  So, when multiple plugins use this module at the same time, I think we cannot know where the `vital-system-logging-log` event comes from.  Is it not a problem?

lambdalisue:
> Why you use of? Python's Logging module doesn't seem to use the name. And Logging.of({name}) seems not to make sense (In addition, it can omit the argument, Logging.of() ). I think it should be renamed.

Python use `get_logger` but it is a bit redundant. Hum... then how about the `Logging.get([{name}])`.

> something like :execute 'doautocmd User vita-system-logging-log-' . vital_version would be handy for future compatibility

>Vital module is embedded in a plugin. So, when multiple plugins use this module at the same time, I think we cannot know where the vital-system-logging-log event comes from. Isn't it a problem?

Then how about the `execute printf('doautocmd User vital-system-logging-log-%s-', vital_version)` (what ujihisa suggest).
In this case, vital in gita call `vital-system-logging-log-gita` and vital in unite call `vital-system-logging-log-unite`. The current vital does not provide the name so some implementation like below will be required.

```vim
function! vital#of(name) abort
  " .. (ry
    throw 'vital: invalid version file: ' . a:name
  endif
  let V = vital#_{substitute(ver[0], '\W', '', 'g')}#new()
  let V.name = a:name
  lockvar V.name
  return V
endfunction
```

rhysd:
> Python use get_logger but it is a bit redundant. Hum... then how about the Logging.get([{name}]).

I agree. But `get_logger` is called per one logger instance and often only one logger exist in a plugin.  So I think redundant name is suitable in this case.

> Then how about the execute printf('doautocmd User vital-system-logging-log-%s', vital_version) (what ujihisa suggest).

It sounds good.  I suggest to add a function to get `vital-system-logging-log-%s` because `vital_version` is for internal use.

tyru:
> Add Vim 7.3 compatible writefile feature to System.File to support Vim 7.3

What you say is {append} flag?
If so, you can use `:redir` instead to append lines to a file.
But be careful. `:redir` isn't nestable.

> Write Changes (is it required to write for a new module?)

It is not required because adding a new module is not destructive change.
But, indeed *changelog* is not for only destructive changes typically.
It is a little bit confusing filename.

autoload/vital/__latest__/System/File.vim
```diff
@@ -230,6 +230,23 @@ else
   endfunction
 endif
 
+" Vim 7.3 compatible writefile function
+function! s:writefile(list, fname, ...) abort
+  let flags = get(a:000, 0, '')
+  if empty(flags)
+    call writefile(a:list, a:fname)
+  elseif v:version > 704 || (v:version == 704 && has('patch503'))
```
thinca: You can check this before defining the function.
Example: https://github.com/vim-jp/vital.vim/blob/master/autoload/vital/__latest__/Vim/Compat.vim#L28-L41

autoload/vital/__latest__/System/File.vim
```diff
@@ -230,6 +230,23 @@ else
...
+    call writefile(a:list, a:fname, flags)
+  elseif flags ==# 'a'
```
thinca: `flags` also takes `"b"` as binary.

autoload/vital/__latest__/System/File.vim
```diff
@@ -230,6 +230,23 @@ else
...
+    let contents = filereadable(a:fname) ? readfile(a:fname) : []
+    call writefile(extend(contents, a:list), a:fname)
+  else
+    throw printf(
+          \ 'vital: System.File.writefile(): "%s" is not supported in this Vim',
```
thinca: The correct prefix is `vital: System.File: `.

autoload/vital/__latest__/System/File.vim
thinca: You can use short version for this: `has('patch-7.4.503')`

mattn:
Most easy way to handle legacy version of vim is:

```vim
if has("patch-7.4.503")
  let s:writefile = function('writefile')
else
  function! s:writefile(list, fname, flags) abort
    if a:flags !~ 'a'
      return writefile(a:list, a:fname, a:flags)
    endif
    let f = tempname()
    let r = writefile(a:list, f, substitute(a:flags, 'a', '', 'g'))
    if has("win32") || has("win64")
      silent! exe "!type ".shellescape(f) ">>".shellescape(a:fname)
    else
      silent! exe "!cat ".shellescape(f) ">>".shellescape(a:fname)
    endif
    call delete(f)
    return r
  endfunction
endif
```

autoload/vital/__latest__/System/File.vim
lambdalisue: Then why they use `s:has_version` at `Compat.vim`? Which one should I use?

autoload/vital/__latest__/System/File.vim
lambdalisue: That's why the function throw an exception if non compatible flags are specified.

autoload/vital/__latest__/System/File.vim
lambdalisue: OK. Should I fix others as well (e.g. https://github.com/vim-jp/vital.vim/pull/305/files#diff-59cc2911b1595d4aadcae22f127f8ce4L229)

autoload/vital/__latest__/System/File.vim
lambdalisue: Hum... I see. lol

lambdalisue:
> What you say is {append} flag?
If so, you can use :redir instead to append lines to a file.
But be careful. :redir isn't nestable.

Yes. That's why I used `readfile` and `writefile` instead.
Well, anyway it seems the version @mattn wrote is the best way.

> It is not required because adding a new module is not destructive change.
But, indeed changelog is not for only destructive changes typically.
It is a little bit confusing filename.

Hum... So I don't have to write changelog, then I can't be bothered to do lol.

autoload/vital/__latest__/System/File.vim
tyru: @lambdalisue See `:help has-patch`.

> The "patch-7.4.237" feature means that the Vim version is 7.5 or
> later, or it is version 7.4 and patch 237 was included.
> **Note that this only works for patch 7.4.237 and later, before that you
> need to use the example above that checks v:version.**

Vim.Compat.has_version() can support this style even before 7.4.237 .

lambdalisue: Got it. Thanks!

autoload/vital/__latest__.vim
```diff
@@ -288,7 +292,7 @@ else
       if a:list[i] ==# a:list[i - 1]
         call remove(a:list, i)
         let i -= 2
-      else
+     else
```
lambdalisue: !

mattn: !

lambdalisue:
あ　き　た

ここまでふくざつなものいらないと思ったので閉じます（笑。
まず外部モジュールで（シンプルなの）作ってできたらPRします。

I just can't be bothered to make this any more while it is way too complicated and prob. nobody need this kind of complicated logger.

ujihisa:
欲しさ

PR #308, opened 2015-07-25
System.File: Escape # and % in Windows

lambdalisue:
With `System.File.open()`, I faced the following exception in Windows

```
E194: No alternate file name to substitute for '#': !start rundll32 url.dll,FileProtocolHandler https://github.com/lambdalisue/vim-gita/blob/hotfix/noshellslash_in_windows/autoload/gita.vim#L2
```

It seems that '#' and '%' required to be escaped due to the [`cmdline-special`](http://vimdoc.sourceforge.net/htmldoc/cmdline.html#cmdline-special), so I escape the two character before passing to command line.

### DIFF
--- a/autoload/vital/__latest__/System/File.vim
+++ b/autoload/vital/__latest__/System/File.vim
@@ -24,7 +24,12 @@ function! s:open(filename) abort "{{{
     if s:need_trans
       let filename = iconv(filename, &encoding, 'char')
     endif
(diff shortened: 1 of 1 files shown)


thinca:
:+1:

lambdalisue:
:smile:

tyru:
:+1:

PR #349, opened 2015-12-09
System.Filepath: Add is_root_directory()

ujihisa:
The function that I've mentioned in Lingr Vim room long ago. I've fixed Windows tests and amended.

This function is based on neoclojure's internal util one.


lambdalisue:
:+1:

PR #401, opened 2016-02-27
Fix plugin_name()

thinca:
When there are two or more plugins that uses vital and they are placed to the
same runtimepath, `plugin_name()` returns wrong name.


Shougo:
This can be merged?

thinca:
Yes.

Shougo:
Merged.

PR #420, opened 2016-03-27
Travis test on mac osx

haya14busa:
We should run test on OSX as well.

I guess we don't need to run vimlint for all platform (it takes about 3min to run vimlint), so I changed .travis.yml to run vimlint only on linux and HEAD env.


thinca:
:+1:

test/System/Filepath.vimspec
```diff
@@ -191,6 +193,9 @@ Describe System.Filepath
 
       Context in case-sensitive system
         It should not ignore case
+          if !is_case_sensitive_file_system
+            Skip "not in case sensitive system, maybe Mac OS?"
```
tyru: or Windows

.travis.yml
```diff
@@ -16,8 +20,11 @@ addons:
       - lua5.1
 
 install:
-  - bash scripts/install-vim.sh
-  - if [ x"$HEAD" = "xyes" ]; then export PATH=$HOME/vim/bin:$PATH; fi
+  - if [ x"$HEAD" = "xyes" ] && [ "$TRAVIS_OS_NAME" != "osx" ]; then bash scripts/install-vim.sh; fi
```
tyru: I don't know .travis.yml well... but this means the following?

> run vimlint only on linux and HEAD env.

(`[ x"$HEAD" = "xyes" ]` wasn't here for `scripts/install-vim.sh`)

test/System/Filepath.vimspec
haya14busa: This line is wrapped with `if !s:is_windows` https://github.com/vim-jp/vital.vim/pull/420/files#diff-8faaef8f6c381bd9962bf5891f98e0b6R163 , so that's wrong.

.travis.yml
haya14busa: scripts/install-vim.sh installs latest vim and HEAD env means running test with latest vim.
I don't know why but before this PR, travis compiles latest vim and didn't use it for non HEAD env.

test/System/Filepath.vimspec
tyru: OK

.travis.yml
tyru: OK

tyru:
LGTM!

haya14busa:
Thank you! :sparkles:

lambdalisue:
:tada:

PR #427, opened 2016-04-29
Appveyor:  Clone depth number is too small

crazymaster:
If you do a lot of commits producing queued builds, git checkout operation following git clone can fail because requested commit is not present in a cloned repository.

ref: https://www.appveyor.com/docs/how-to/repository-shallow-clone#setting-depth-of-git-clone-command

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 version: '{build}'
-clone_depth: 1
+clone_depth: 10
 environment:
   matrix:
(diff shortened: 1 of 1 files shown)


haya14busa:
+1

It's not directly related with this PR, but should we add an corresponding option in travis ci?
https://docs.travis-ci.com/user/customizing-the-build/#Git-Clone-Depth
```
git:
  depth: 10
```

PR #437, opened 2016-06-04
Data.Closure: Fix for Partial

thinca:
In latest Vim, `dict.func` returns a Partial.  And they are compared including the internals.  We don't expect it.


tyru:
LGTM

PR #453, opened 2016-09-26
Fix doc (vital-option_parser.txt)

ryunix:
0. Remove invalid option (`{short-name}` is already deleted)
0. Alignment


ryunix:
Thanks :)

PR #464, opened 2016-12-15
Introduce reviewdog.yml

haya14busa:
I released reviewdog v0.9.3. https://github.com/haya14busa/reviewdog/releases/tag/0.9.3 and 
I bumped up reviewdog version in dockerfile-lint-vital.vim https://github.com/haya14busa/dockerfile-lint-vital.vim

As of 0.9.3, reviewdog can also be controlled via the reviewdog.yml configuration file instead of "-f" or "-efm" arguments.

With reviewdog.yml, you can run the same commands both in CI service and **local environment**.

example:

```
$ reviewdog -diff="git diff master"
autoload/vital/__vital__/Data/String.vim:9:11: [vint] Prefer single quoted strings (see Google VimScript Style Guide (Strings))
autoload/vital/__vital__/Data/String.vim:9:7: [vimlint] EVL102: unused variable `l:a`
```

reviewdog.yml feature is somewhat experimental and there are lots of room for improvement, but I'll take care the compatibility issues.
So, I'm really glad if you include this changes.


autoload/vital/__vital__/Data/String.vim
```diff
@@ -6,6 +6,7 @@ set cpo&vim
 function! s:_vital_loaded(V) abort
   let s:V = a:V
   let s:L = s:V.import('Data.List')
+  let a = "double quote"
```
haya14busa: ok :+!:

haya14busa:
rebased. Please take a look.

Issue #474, opened 2017-01-16
Please describe what vital.vim provides *with a paragraph*


umireon:
**As a newbie on vital.vim**

I cannot understand what vital.vim provides and aims at.
I know [the project page](https://github.com/vim-jp/vital.vim#readme) has a *comprehensive* list of modules but I don't want to *check it out* such a long and **uninteresting** table, which is **only** repeating the same words in the same line.
In addition, [the brief description field](https://github.com/vim-jp/vital.vim#readme) and [the introduction section](#vitalvim--) gave me no information but the word *comprehensive* and the idea “Ah! RUBY and JS can be combined!”

Can you give me a description as written in the first sentence of [Boost on Wikipedia](https://en.wikipedia.org/wiki/Boost_(C%2B%2B_libraries)), which gives me a clear concept what Boost provides?

tyru:
At the beginning of vital.vim, vital.vim is the library to just cover the poor Vim script capability.
But in Vim script, using external library brings another problems:

* Does the library place to single repository or not?
* If the answer is "Yes"
  * Updating the library is tedious
  * Checking destructing changes are super tedious
* If the answer is "No"
  * User must install it manually
  * There is no version locking system in Vim script

The above problems are only when one library depends no other libraries.
If it depends other libraries or when using multiple libraries, the problems become more complicated. 
vital.vim solves the above problems.

umireon:
> vital.vim solves the above problems.
By what means? By concatenating the libraries on the internet into one file as [browserify](http://qiita.com/cognitom/items/4c63969b5085c90639d4) does?
If so, where is the software repository like [npm](https://www.npmjs.com/search?q=random)?

tyru:
vital.vim solves the problems by:

* vital.vim **embed** libraries into developer's repository so user doesn't have to install manually, and also there is no need to care about a version of libraries.
* Vital "modules" can depend other "modules".
* When embedding modules, vital resolves the dependencies and install only necessary modules.
* If there are already embedded modules in repositories, vital updates existing modules. then ...
  * Developer can also specify adding modules / removing modules.
  * Vital shows destructive changes if embedding modules contain them.

umireon:
Thank you for the detailed technical description, @tyru.
I got your point that vital.vim is a module bundler in Vim.
So, how can I bundle vimproc to my plugin with vital.vim?

tyru:
vital.vim can bundle only "modules."
vimproc is not a vital module. so you can't.
[...]

PR #475, opened 2017-01-18
CI: Prefer GitHub tarballs to shallow cloning

umireon:
Downloading a GitHub tarball is a straightforward way.
`git` command is too complex for tasks such as downloading an external software and simple commands will make builds more robust.

## On the pipeline
* `curl -fsSL`: Downloads silently with following the redirect and writes the content to stdout, but shows the error and fails when some errors occur.
  `curl` is also available on Appveyor because MinGW in the git distribution contains it.
* `tar -xz -C dir`: Extract the gzipped tarball from stdin under `dir`.
* `7z x -y -si -so -tgzip`: Extract the gzipped file from stdin to stdout with answering yes to all the questions. Equivalent to `zcat`.
* `7z x -y -si -odir -ttar`: Extract the uncompressed tarball from stdin under `dir` with answering yes to all the questions. Equivalent to `tar -x -C dir`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ sudo: false
 git:
   depth: 10
 
+env:
+  global:
+    - VIMPROC_VERSION=master
+    - VIM_THEMIS_VERSION=v1.5.2.1
+
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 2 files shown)


tyru:
ベンチマークが欲しいです。

umireon:
I edited the description to focus on the simplicity.
Performance should not be a point because the current build process is sufficiently fast.
Using GitHub tarballs seems to make builds a bit faster by 10 seconds, though.

tyru:
僕個人の意見なのですが、正直 git の代わりに curl を使うことでシンプルになると言われても、それがどのような点で改善になるのかが分かっていません。
「CI の完了までの時間が短縮できる」なら分かりますが、高速化の話ならベンチマークを付けてほしい、というのが僕個人の意見です。

haya14busa:
I agree with @tyru. git command is not complex. In addition, it runs on CI server, so most of developers doesn't have to care about it.

vital.vim currently uses dependencies with tagged release, but we don't have clear agreement AFAIK, so we may depends external git repositories without tagged release, using latest HEAD, branch or something... which we cannot use curl to get their code.

umireon:
@haya14busa Thank you for your sincere comments. There are a lot for me to be learned.

> git command is not complex

`git clone` is indeed complex in the aspect what it actually does.
Particular to shallow cloning, it gets refs or packed-refs and parse them and get the hashes of desired commits and gets all the git objects which are needed and uncompress them into the working directory.
This compositeness increases the number of HTTP requests and the probability of build failure because the network connectivity of CI servers is often unstable.

`curl` just makes one or two HTTPS requests and `tar` is tar. 

> In addition, it runs on CI server, so most of developers doesn't have to care about it.

So do `curl` and `tar`. Machines everywhere have them.

> which we cannot use curl to get their code.

I'm afraid that your statement is not true.
Here we can have https://github.com/thinca/vim-themis/archive/master.tar.gz and https://github.com/thinca/vim-themis/archive/176868b6784d5b7281898c4cfb702d1da5f65b03.tar.gz

k-takata:
（vitalにはほとんど関わっていない部外者の意見ですが、）アーカイブ解凍後のディレクトリ名がバージョン（タグ名）に依存しており、バージョンを変更する際には複数箇所修正が必要となるため、かえって複雑になっていると感じました。

umireon:
@k-takata それは`--strip-components 1`で解決可能です。

haya14busa:
> git clone is indeed complex in the aspect what it actually does.

I understand git clone does complex things internally. I mean, in terms of maintainability of CI script, it's not difficult. In addition, we are used to git-clone-from-git-repository workflow.

To be honest, curl & tar & 7z  look more difficult for me than git clone workflow, especially 7z...
You explained how they works in the top of comments in this pull-request but... we don't need detail explanation at the first place for git clone workflow.

If you insist that curl & tar workflow are fast and robust, I understand it's better than git for these points and it's might be worth considering to accept the change. However, you changed the focus to simplicity and I doubt it.

I appreciate your contributions, especially I like #474, but I cannot approve this pull-request.

If you still want this pull-request included, I think you might be better to focus on performance or robustness, not subjective simplicity or something...
And I leave it to others to accept this pull-request.

To be honest, I think both are fine and it doesn't make big differences.

umireon:
@haya14busa I appreciate your understandings on my points and I'm greatly convinced.
Please note that I DO have been focusing on the robustness and NEVER on subjective simplicity.
Anyway, feel free to close this PR. This PR revealed as just a preference.

umireon:
Ah! So sorry... 🙇 
Apparently, I was dazzled with the word of simplicity and robustness.
I have wanted to make a strong focus on the robustness, but I failed to choose appropriate words.
It's all my fault...

thinca:
Many CI services use Git to obtain project code.
Git is a standard tool around CI, so Git is not complex for this purpose.

PR #477, opened 2017-01-19
System.Filepath: Case-insensitive only on Windows

umireon:
This PR fixes the problem that `Filepath.contains` mistakenly recognizes a POSIX system as a case-torelant system when the library is placed on a case-insensitive filesystem such as [osxfs in Docker](https://docs.docker.com/docker-for-mac/osxfs/).

## Rationale
POSIX paths (on Linux, BSD, and macOS) are case-sensitive in nature, and individual filesystems can have different settings.
Windows paths are case-insensitive.

### DIFF
--- a/autoload/vital/__vital__/System/Filepath.vim
+++ b/autoload/vital/__vital__/System/Filepath.vim
@@ -12,7 +12,7 @@ let s:is_cygwin = has('win32unix')
 let s:is_mac = !s:is_windows && !s:is_cygwin
       \ && (has('mac') || has('macunix') || has('gui_macvim') ||
       \   (!isdirectory('/proc') && executable('sw_vers')))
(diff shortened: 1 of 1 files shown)


lambdalisue:
Case sensitivity and OS is a different topic (e.g. Mac OS X use case insensitive in default). So I prefer a better modified version of the current implementation.

haya14busa:
> I prefer a better modified version of the current implementation.

What is the problem of current implementation in the first place???

If `Filepath.VIM` is readable, then the file system is case-insensitive. It's simple and works as expected in every file system as far as I understand.

haya14busa:
Oh, I didn't read the top comment carefully. So it doesn't work for shared file system? 
Anyway, we cannot accept this fix as lambdalisue commented I think

ujihisa:
I'm checking through all existing pull requests. Should I leave this pull request open to let people discuss more, or close this?

PR #483, opened 2017-02-17
Fix tags in introduction document

rhysd:
I noticed that `:h vital.vim` doesn't show introduction of vital.vim. [Link to `vital.vim`](https://github.com/vim-jp/vital.vim/blob/06f830e2c10ee26f94b5665c73da72c6c9b422c5/doc/vital.txt#L40) even exists... So I added it.

(Note that `:h vital` also can't lead users to introduction page because tag is case sensitive. Should we change `Vital` tag to `vital`?)

### DIFF
--- a/doc/vital.txt
+++ b/doc/vital.txt
@@ -16,8 +16,8 @@ LINKS				|Vital-links|
 ==============================================================================
 INTRODUCTION				*Vital-introduction*
 
-*Vital* is like a plugin which has both aspects of Bundler and jQuery at the
-same time.
+*vital* (or *vital.vim* ) is like a plugin which has both aspects of Bundler and
+jQuery at the same time.
 
  Bundler: http://gembundler.com/
  jQuery:  http://jquery.com/


lambdalisue:
LGTM :+1: 

> (Note that :h vital also can't lead users to introduction page because tag is case sensitive. Should we change Vital tag to vital?)

Agree. A capital word is not useful so I think we should change **or** add `vital` while `Vital` has been used so long time.

rhysd:
So, let's say to change `Vital` to `vital`!

ryunix:
桁数が78を超えてしまっていますね。
jQuery の手前で改行をするのが適切かと思いましたが、いかがでしょうか？

rhysd:
@ryunix ありがとうございます，見落としてました…

rhysd:
I'll merge this tomorrow evening. If anyone has some comments, please leave them.

doc/vital.txt
```diff
@@ -16,8 +16,8 @@ LINKS				|Vital-links|
 ==============================================================================
 INTRODUCTION				*Vital-introduction*
 
-*Vital* is like a plugin which has both aspects of Bundler and jQuery at the
-same time.
+*vital* or *vital.vim* is like a plugin which has both aspects of Bundler and
```
haya14busa: How about removing "*vital*" and just leaving "vital.vim" only?

I feel "or" is a bit redundant and `:h vital` can work without "*vital*" tag.

rhysd: As long as tried, `:help vital` hits `vital#of()` rather than `vital.vim`... So I think `vital` tag is needed here.

haya14busa: Hm... Thanks. I agree with you.

it's a nit comment after all, but then, how about `*vital* (*vital.vim*) is ...`?

"or" seems a normal sentence, so I read "or *vital.vim*" silently... but just either "vital" or "vital.vim" should be fine in this sentence.

rhysd: I feel your idea looks good for me also. If no other specific comments about this, I'll adopt it :+1:

rhysd: Ah, `*vital* (*vital.vim*)` doesn't highlight `*vital.vim*`. We need to use `*vital* (  *vital.vim* )`...

tyru: hmm... How about writing `*vital*` and `*vital.vim*` in line 1 in the doc?
it gets the sentence more natural.

rhysd: It seems that header can contain one tag for the list in `:h local-additions`. So it's unavailable...

rhysd:
I found `*vital* (or *vital.vim*)` works. How about this?

doc/vital.txt
tyru: okay...

tyru:
@rhysd LGTM!

rhysd:
I misunderstood the highlight. Space just after a tag is necessary, so I added workaround (adding a space after the last asterisk).

PR #485, opened 2017-03-03
DateTime: Fix timezone on Windows 64 bit + Vim 32 bit

bimlas:
Tests failed on Windows 64 because `REG_DWORD` is 32 bit long number,
but integers are stored on 64 bit, thus it have to be expanded to 64 bit
to get the right value (for negative numbers).


bimlas:
Here's my [output of Themis before this patch](https://gist.github.com/bimlas/fabc97e6df5ce3026916fccf9f76355f).

rhysd:
Thank you for the patch. I added an author of DateTime lib to reviews.

PR #488, opened 2017-03-15
Fix tests on Mac OS X in Travis

lambdalisue:
It seems `brew upgrade` is required to upgrade the version of Python on Mac OS X in Travis


lambdalisue:
テスト通り次第マージします

lambdalisue:
Thanks!

PR #495, opened 2017-04-16
Data.Dict: Add .lookup()

aiya000:
Haskell like .lookup() 😄

### DIFF
--- a/autoload/vital/__vital__/Data/Dict.vim
+++ b/autoload/vital/__vital__/Data/Dict.vim
@@ -104,7 +104,7 @@ function! s:foldr(f, init, dict) abort
   return s:_foldl(a:f, a:init, reverse(items(a:dict)))
 endfunction
 
-function! s:lookup(key, dict) abort
+function! s:lookup(dict, key) abort
 	return has_key(a:dict, a:key)
 	\        ? s:Option.some(a:dict[a:key])
 	\        : s:Option.none()
--- a/autoload/vital/__vital__/Data/Dict.vim
+++ b/autoload/vital/__vital__/Data/Dict.vim
@@ -3,6 +3,15 @@
 let s:save_cpo = &cpo
 set cpo&vim
 
+function! s:_vital_loaded(V) abort
+  let s:V = a:V
+  let s:Option = s:V.import('Data.Optional')
+endfunction
+
+function! s:_vital_depends() abort
+  return ['Data.Optional']
+endfunction
+
 " Makes a dict from keys and values
 function! s:make(keys, values, ...) abort
   let dict = {}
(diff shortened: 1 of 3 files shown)


doc/vital/Data/Dict.txt
```diff
@@ -113,5 +113,17 @@ foldr({f}, {init}, {dict})		*Vital.Data.Dict.foldr()*
 	TODO
 	Behaves like Haskell's Data.Map.Lazy.foldrWithKey().
 
+
+lookup({key}, {dict})			*Vital.Data.Dict.lookup()*
+	If the dict has the key, returns Optional.some({dict}[{key}]).
+	If the dict hasn't the key, returns Optional.none().
```
ujihisa: s/hasn't/doesn't have`

ujihisa: Optionalのdocへのlinkもほしい

ujihisa:
👍 :sushi: looks good except for a minor typo.

It may be also helpful if there's a doc in Optional side to mention about this Data.Dict.lookup.

haya14busa:
I think we should add lookup method to Data.Optional.

I prefer fewer dependencies, especially if the dependent module is rarely used.
Data.Optional is completely different paradigm too. I think almost all Data.Dict users doesn't use lookup method.

Adding dependencies to Data.Dict means all Vim plugins which depends on Data.Dict will include and load Data.Optional too, even if they don't use lookup method.

autoload/vital/__vital__/Data/Dict.vim
```diff
@@ -95,6 +104,12 @@ function! s:foldr(f, init, dict) abort
   return s:_foldl(a:f, a:init, reverse(items(a:dict)))
 endfunction
 
+function! s:lookup(key, dict) abort
```
haya14busa: I think the order of arguments should be dict -> key to be consistent with builtin has_key() function.

I suppose Haskell uses key -> dict order, but it's Vim script after all.

aiya000:
I finished it except namespace locating :+1:

mmm,
How do you think to create Data.Optional.Dict ?

rhysd:
Just a question: Why don't you use `get(dict, key, default)`?

rhysd:
My suggestion: Move this to `s:Optional.lookup(collection, elem)` where `collection` is a list or dict. I think supporting both list and dict would be more useful.

aiya000:
!!

I forgot `get()` :joy:  
`get()` satisfies my use case...

aiya000:
If nobody needs this, I think this should be removed.

If we keep this, its idea is good.

ujihisa:
The potential benefit of returning Some instead of specifying return value by `default` of `get` is chaining, but yeah it's Vim script.

I close this and if somebody want this feature that person can reopen again. Probably not as `Data.Dict.lookup` but as a function in optional side.

aiya000:
@ujihisa Thanks 👍

PR #496, opened 2017-04-16
Data.Optional: Add .map()

aiya000:
For simple API 🤔  
(I think 'map' is most popular name 😄 )

### DIFF
--- a/autoload/vital/__vital__/Data/Optional.vim
+++ b/autoload/vital/__vital__/Data/Optional.vim
@@ -86,6 +86,10 @@ function! s:apply(f, ...) abort
   return s:some(call(a:f, map(copy(a:000), 'v:val[0]')))
 endfunction
 
+function! s:map(f, x) abort
+	return s:apply(a:f, a:x)
+endfunction
+
 function! s:bind(f, ...) abort
   if !s:_valid_args(a:000)
     return s:none()
--- a/autoload/vital/__vital__/Data/Optional.vim
+++ b/autoload/vital/__vital__/Data/Optional.vim
@@ -87,7 +87,7 @@ function! s:apply(f, ...) abort
 endfunction
 
 function! s:map(x, f) abort
-	return s:apply(a:f, a:x)
+  return s:apply(a:f, a:x)
 endfunction
 
 function! s:bind(f, ...) abort
--- a/autoload/vital/__vital__/Data/Optional.vim
+++ b/autoload/vital/__vital__/Data/Optional.vim
@@ -86,7 +86,7 @@ function! s:apply(f, ...) abort
   return s:some(call(a:f, map(copy(a:000), 'v:val[0]')))
 endfunction
 
-function! s:map(f, x) abort
+function! s:map(x, f) abort
 	return s:apply(a:f, a:x)
 endfunction
 
--- a/doc/vital/Data/Optional.txt
+++ b/doc/vital/Data/Optional.txt
@@ -130,8 +130,17 @@ apply({func}, {args}...)		*Vital.Data.Optional.apply()*
 	call O.apply(function('Update'), s:cache, O.some(new_value))
 <
 map({optional}, {func})			*Vital.Data.Optional.map()*
-	A swap case of |Vital.Data.Optional.apply()| but this function takes
-	only an argument.
+	Maps content of {optional} by predicate {func}. If content of
+	{optional} is none, then {func} won't be invoked and none will be
+	returned.
+>
+	function! Succ(x)
+	  return a:x + 1
+	endfunction
+
+	echo O.map(O.some(1), function('Succ')) " returns O.some(2)
+	echo O.map(O.none(), function('Succ'))  " returns O.none()
+>
 
 bind({func}, {args}...)			*Vital.Data.Optional.bind()*
 	Applies {args} to {func} if all of {args} are valid optional value.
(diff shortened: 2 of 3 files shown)


test/Data/Optional.vimspec
```diff
@@ -163,6 +163,13 @@ Describe Data.Optional
     End
   End
 
+  Describe .map()
+    It can take only an argument
+      Assert Equal(O.map(O.some(1), {x -> x + 1}), O.some(2))
```
rhysd: Lambda expression can only be used for Vim8. So it fails with Vim 7.3 or 7.4. I think we need to guard by Vim's version here because vital.vim is tested with 7.4 also.

test/Data/Optional.vimspec
```diff
@@ -163,6 +163,13 @@ Describe Data.Optional
...
+      Throws /^Vim(return):E118:/ O.map(O.some(1), {x -> x + 1}, O.some(2))
```
rhysd: I think this `Throws` test is not necessary because it's obvious by its definition.

test/Data/Optional.vimspec
```diff
@@ -163,6 +163,13 @@ Describe Data.Optional
     End
   End
 
+  Describe .map()
+    It can take only an argument
```
rhysd: I think we should test basic behavior of `.map()` here (like `It maps content of Optional`)

autoload/vital/__vital__/Data/Optional.vim
```diff
@@ -86,6 +86,10 @@ function! s:apply(f, ...) abort
   return s:some(call(a:f, map(copy(a:000), 'v:val[0]')))
 endfunction
 
+function! s:map(x, f) abort
+	return s:apply(a:f, a:x)
```
rhysd: Indentation

doc/vital/Data/Optional.txt
```diff
@@ -129,6 +129,9 @@ apply({func}, {args}...)		*Vital.Data.Optional.apply()*
 	" a valid value.
 	call O.apply(function('Update'), s:cache, O.some(new_value))
 <
+map({optional}, {func})			*Vital.Data.Optional.map()*
+	A swap case of |Vital.Data.Optional.apply()| but this function takes
```
rhysd: What is 'swap case' here?

rhysd: And what does 'this function' indicates? I think adding an example makes more descriptive.

test/Data/Optional.vimspec
```diff
@@ -163,6 +163,13 @@ Describe Data.Optional
...
+      Assert Equal(O.map(O.some(1), {x -> x + 1}), O.some(2))
+      Throws /^Vim(return):E118:/ O.map(O.some(1), {x -> x + 1}, O.some(2))
+    End
```
rhysd: Test case for passing `O.none` to the function should be added.

test/Data/Optional.vimspec
rhysd: Or replacing the lambda with normal funcref is also good (it'd be easier).

doc/vital/Data/Optional.txt
aiya000: This 'swap case' means...
.apply() takes 

- first argument: func
- second argument: optional

.map() reverses it !

Are there better phrases more than this phrase ?

aiya000: mmm, thanks!

I didn't think about {func}.  
I'll fix it! :+1:

rhysd: AFAIK, 'swap case' is not a common words so we should avoid them.

Googled: https://www.google.co.jp/search?q=%22swap+case%22&oq=%22swap+case%22&aqs=chrome..69i57j0l5.5158j0j7&sourceid=chrome&ie=UTF-8

IMO, we can describe the *behavior* of `map()` function. The difference between `apply()`'s parameters and `map()`'s ones is not so important. In general, user reads documentation to know the behavior of API. So simply describe 'Maps content of {optional} by predicate {func}. If content of {optional} is none, then {func} won't be invoked and none will be returned.' Adding an example code would be more appreciated.

test/Data/Optional.vimspec
aiya000: Thanks !  
I did it 😄

test/Data/Optional.vimspec
aiya000: I certainly think so too!  
I fixed it !

aiya000:
I finished it 👍  
I used its word of document, thanks !

test/Data/Optional.vimspec
```diff
@@ -163,6 +163,34 @@ Describe Data.Optional
...
+        return a:x + 1
+      endfunction
+
+      function! Append(x)
+        return 'It is ' . a:x 
+      endfunction
+
+      function! SayPoi(x)
+        return {'say': a:x.say . 'poi'}
+      endfunction
```
thinca: Please delete global functions in after block or use script local function.

aiya000: I finished it :smile:  
(TestFunc() in .apply() test isn't removed, is it OK ?)

thinca: Thanks!

> TestFunc()

Oh, it should be removed. However, it should be done with other PR 😎

aiya000:
Thanks!

PR #517, opened 2017-05-22
Data.Optional: s/end/endfunction/

ujihisa:
* Vim does not have `end`


ujihisa:
@rhysd 今日の https://connpass.com/event/56803/ で発表予定

PR #529, opened 2017-07-05
CI: Use misspell binary

haya14busa:
Building misspell might use lots of resource and be killed.
https://github.com/vim-jp/vital.vim/pull/513#issuecomment-312435191

ref: https://github.com/client9/misspell#install


.drone.yml
```diff
@@ -8,7 +8,7 @@ build:
       - REVIEWDOG_GITHUB_API_TOKEN=$$REVIEWDOG_GITHUB_API_TOKEN
     commands:
       - go get -d -v ./scripts
-      - go get -u github.com/client9/misspell/cmd/misspell
+      - curl -L https://git.io/misspell | sh
```
rhysd: From [misspell's installation document](https://github.com/client9/misspell#install),

> Both will install as ./bin/misspell.

So we might need to specify `misspell` executable with relative path in `reviewdog.yml`.

autoload/vital.vim
```diff
@@ -10,3 +10,5 @@ function! vital#of(name) abort
   endif
   return vital#_{substitute(ver[0], '\W', '', 'g')}#new()
 endfunction
+
+let g:exsiting_variable = 1
```
haya14busa: :bird: :+1:

.drone.yml
haya14busa: Oops, thank you for the catch!

haya14busa:
Fixed. Please take another look.

PR #532, opened 2017-07-21
Process: wrap str as string

josephholsten:
discovered via https://github.com/tyru/open-browser.vim/pull/98


tyru:
@josephholsten Thanks! LGTM.

PR #541, opened 2017-10-08
Fix ruby build

k-takata:
Recently, Travis-CI switched to Ubuntu trusty. At the same time, rvm was introduced and it caused a link error when building vim with if_ruby.
Disable rvm and use the system ruby to fix the problem.

Actually, if_ruby might not be needed...

Related: #540


k-takata:
Yay, I think this is my first commit to vital.vim!

PR #548, opened 2017-11-07
System.Filepath: doc: add example to path_separator()

tyru:
Close #545 

Original PR (#545) which introduce `path_separate()` function was rejected
because `split(path, path_separator())` was enough.

However, `path_separator()` lacks its example in the document.


k-takata:
I think the example should be:

```vim
	split('/usr/bin:/bin', s:F.path_separator()) == ['/usr/bin', '/bin']
```

or something similar. Your example is for `separator()`, not `path_separator()`.

tyru:
oops, I mistaken.
Thanks.

thinca:
oops, I missed too.

PR #553, opened 2017-11-23
vitalizer: Wrap function() to ignore error when creating module function list

### LINKED ISSUE #537: rshift32 import issue with Bitwise
I'm seeing an issue when running my tests on travis where `Bitwise.rshift32` (and presumably `Bitwise.lshift32`) isn't defined when the vital module is imported.

```
./vim-themis/bin/themis --reporter dot -r tests

Error occurred in script loading.
function <SNR>21_import(name, ...) abort dict  (~/build/nicwest/vim-generate/autoload/vital/generate.vim)
function <SNR>21__import(name) abort dict  Line:18  (~/build/nicwest/vim-generate/autoload/vital/generate.vim)
 18:       throw 'vital: fail to call ._vital_loaded(): ' . v:exception
vital: fail to call ._vital_loaded(): Vim(return):E700: Unknown function: s:rshift32
make: *** [test] Error 1
```

https://travis-ci.org/nicwest/vim-generate/jobs/268990406
https://github.com/nicwest/vim-generate/tree/13dbcfe0b3097d29c28080d06f2de195b5c039c4

I fixed it in my repo with this change:

https://github.com/nicwest/vim-generate/commit/75b7214032fea99b5b9f8f5d8bf343858701f9c9

I don't know if there is a better way to go about this, or maybe it's an artifact of the way I'm using vital?

haya14busa:
The function list is generated on Vitalize and `#import()` func is called in users environment, some function might not be defined as script local function and `function('s:{funcname}')` will throw errors.

Fix #537


haya14busa:
I updated the commit message to describe why we switched to use try-catch when calling `#import()` function.

PR #557, opened 2017-12-16
vitalizer: Avoid the bug of execute()

thinca:
`execute()` has a bug.  This PR avoid it.

See also(Japanese): https://github.com/vim-jp/issues/issues/1129

### DIFF
--- a/autoload/vital/vital.vim
+++ b/autoload/vital/vital.vim
@@ -257,20 +257,20 @@ function! s:_sid(path, filter_pattern) abort
   return 0
 endfunction
 
-" A bug of execute() is fixed in Vim 8.0.0264
-if has('patch-8.0.0264')
-  let s:_execute = function('execute')
-else
-  function! s:_execute(cmd) abort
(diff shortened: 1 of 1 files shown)


autoload/vital/vital.vim
```diff
@@ -257,10 +257,16 @@ function! s:_sid(path, filter_pattern) abort
   return 0
 endfunction
 
+
+" There is still bug in execute().
+" execute() returns empty string when it is called in
+" completion function of user defined ex command.
+" We will re-enable following codes when the bug is fixed.
```
rhysd: If you already reported this bug to vim-jp, could you please add the URL to track the issue?

autoload/vital/vital.vim
```diff
@@ -257,10 +257,16 @@ function! s:_sid(path, filter_pattern) abort
...
+" ---
 " A bug of execute() is fixed in Vim 8.0.0264
```
rhysd: I think this comment should be removed because `execute()` is still broken.

PR #558, opened 2017-12-16
Remove obsolete file "autoload/vital.vim"

thinca:
`autoload/vital.vim` is obsolete and it should not be used.
In addition, this file conflicts between each Vim plugins.

I think this file is no longer needed.  So remove this.


Changes
```diff
@@ -1,3 +1,7 @@
+c680503cd8e48f7b1c21903ed14ffe75cb56c3f4
+	Modules: *
+	The obsolete file "autoload/vital.vim" is no longer installed
+	and "vital#of()" is no longer available.
```
rhysd: How about writing an alternative to `vital#of`? (i.e. "vital#of()" is no longer available. Please use vital#{plugin}#new() instead.)

thinca: I was thinking the same thing. 😃

PR #559, opened 2017-12-16
codecov: Try to allow any coverage drop

rhysd:
Only 0.01% coverage drop is reported as CI failure by codecov. I want codecov to be success always because coverage tool (covimerage) is still not so stable and the coverage is just a rough indication of test quality.


rhysd:
codecov ムズすぎる

PR #564, opened 2017-12-21
Vim.Type: Add is_comparable() function

thinca:
This is useful when user want to compare any two values.

### DIFF
--- a/autoload/vital/__vital__/Vim/Type.vim
+++ b/autoload/vital/__vital__/Vim/Type.vim
@@ -47,3 +47,37 @@ function! s:is_predicate(value) abort
   return t == s:types.number || t == s:types.string ||
   \ t == s:types.bool || t == s:types.none
 endfunction
+
+function! s:is_comparable(value1, value2) abort
+  if !exists('s:is_comparable_cache')
+    let s:is_comparable_cache = s:_make_is_comparable_cache()
+  endif
+  return s:is_comparable_cache[type(a:value1)][type(a:value2)]
+endfunction
+
+function! s:_make_is_comparable_cache() abort
+  let vals = [
+  \   0, '', function('type'), [], {}, 0.0,
+  \   get(v:, 'false'),
+  \   get(v:, 'null'),
+  \   exists('*test_null_job') ? test_null_job() : 0,
+  \   exists('*test_null_channel') ? test_null_channel() : 0,
+  \ ]
+
+  let result = []
+  for l:V1 in vals
+    let result_V1 = []
+    let result += [result_V1]
+    for l:V2 in vals
+      try
+        let _ = V1 == V2
+        let result_V1 += [1]
+      catch
+        let result_V1 += [0]
+      endtry
+      unlet V2
+    endfor
+    unlet V1
+  endfor
+  return result
+endfunction
--- a/doc/vital/Vim/Type.txt
+++ b/doc/vital/Vim/Type.txt
@@ -99,6 +99,10 @@ is_predicate({value})			*Vital.Vim.Type.is_predicate()*
 	  return v:val ? 'true' : 'false'
 	endfunction
 
+is_comparable({value1}, {value2})	*Vital.Vim.Type.is_comparable()*
+	Returns TRUE if {value1} and {value2} can be compared by |expr-==|,
+	FALSE otherwise.
+
 
 
 ==============================================================================
--- a/test/Vim/Type.vimspec
+++ b/test/Vim/Type.vimspec
@@ -145,4 +145,52 @@ Describe Vim.Type
       End
     End
   End
+
+  Describe .is_comparable()
+    Before each
+      let vals = [0, '', function('type'), [], {}, 0.0]
+      if 800 <= v:version
+        let vals += [v:false, v:null, test_null_job(), test_null_channel()]
+      endif
+
+      let true_group = []
+      let false_group = []
+      for l:V1 in vals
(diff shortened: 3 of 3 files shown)


tyru:
oops... sorry.
I missed the test result seems failing...
But LGTM if the results become success.

thinca:
fmm... `number` and `funcref` can be compared in Vim 8.0, but it can not in Vim 7.4 🤔 
Should I change the behavior in each versions...?

tyru:
> Should I change the behavior in each versions...?

I think so.

Well, sorry to say now, I wonder this function is useful.
Because you can compare two values without errors as follows, how about this?

```vim
try
  return V1 ==# V2
catch
  return 0
endtry
```

autoload/vital/__vital__/Vim/Type.vim
```diff
@@ -47,3 +47,18 @@ function! s:is_predicate(value) abort
...
+  let tname2 = s:type_names[t2]
+  if tname1 ==# 'list' || tname1 ==# 'dict' ||
+  \   tname2 ==# 'list' || tname2 ==# 'dict'
+    return tname1 ==# tname2
+  endif
+  if (tname2 ==# 'job' || tname2 ==# 'channel') &&
+  \   (tname1 ==# 'number' || tname1 ==# 'float')
+    return 0
+  endif
+  return 1
```
rhysd: Possible another implementation:

```vim
function! s:is_comparable(v1, v2) abort
  try
    a:v1 == a:v2
    return 1
  catch
    return 0
  endtry
endfunction
```

rhysd:
ah, sorry for the duplication. My browser shows outdated cache of this page and I did not notice the comment...

thinca:
Hmm... I think `try` should not be used for normal pass.

thinca:
I tried to compare performance between the current implementation and using try-catch by following code.

```
    It performance
      let v = repeat(vals, 30)

      let t = reltime()
      for V1 in v
        for V2 in v
          call T.is_comparable(V1, V2)
        endfor
      endfor
      let t = reltime(t)
      call themis#log(reltimestr(t))

      let t = reltime()
      for V1 in v
        for V2 in v
          call T.is_comparable2(V1, V2)
        endfor
      endfor
      let t = reltime(t)
      call themis#log(reltimestr(t))
    End
```

```
function! s:is_comparable2(value1, value2) abort
  try
    let _ = a:value1 == a:value2
    return 1
  catch
    return 0
  endtry
endfunction
```

As result, using try-catch is faster.

```
  2.678315
  1.655016
```

I imagined `try-catch` to be slow, but it was different.
I'll use `try-catch` here for performance.  Thanks @tyru and @rhysd !

thinca:
I force-pushed and the old implementation was lost 🙇 

I put here the old (called `current` in the previous comment) implementation for the someday.

```vim
function! s:is_comparable(value1, value2) abort
  let [t1, t2] = sort([type(a:value1), type(a:value2)])
  let tname1 = s:type_names[t1]
  let tname2 = s:type_names[t2]
  if tname1 ==# 'list' || tname1 ==# 'dict' ||
  \   tname2 ==# 'list' || tname2 ==# 'dict'
    return tname1 ==# tname2
  endif
  if (tname2 ==# 'job' || tname2 ==# 'channel') &&
  \   (tname1 ==# 'number' || tname1 ==# 'float')
    return 0
  endif
  return 1
endfunction
```

thinca:
All tests passed 🎉

k-takata:
I think `==` can be slow when `value1` and `value2` are both long lists.

thinca:
I agree...

k-takata:
Profiled: https://gist.github.com/k-takata/7a692bceed281fce23e56d4d179c52b7
My expectation was wrong. Comparing long lists was not so slow. However, `[] == {}` was slow. I don't know why.

thinca:
New approach: Cache the results in first time and refernce it later.
This is fastest when function is called many time.
https://wandbox.org/permlink/LXBI7VKSeiQKJaDJ

But, first time make be slow a little.  I think this is acceptable.
https://wandbox.org/permlink/P7iAywfq0hqCpVgf

autoload/vital/__vital__/Vim/Type.vim
```diff
@@ -47,3 +47,37 @@ function! s:is_predicate(value) abort
   return t == s:types.number || t == s:types.string ||
   \ t == s:types.bool || t == s:types.none
 endfunction
+
+function! s:is_comparable(value1, value2) abort
+  if !exists('s:is_comparable_cache')
+    let s:is_comparable_cache = s:_make_is_comparable_cache()
+  endif
+  return s:is_comparable_cache[type(a:value1)][type(a:value2)]
```
tyru: LGTM. But I have one **suggestion**:
* To make cache faster, and less memory footprint, how about constraining `t1 <= t2`?

```vim
let [t1, t2] = [type(a:value1), type(a:value2)]
return s:is_comparable_cache[t1 <= t2 ? t1 : t2][t1 <= t2 ? t2 : t1]
```

and `s:_make_is_comparable_cache()` makes only the necessary results.

thinca: It seems maybe make slow to access cache.
Could you measure performance of making cache and accessing to cache in current code and suggested code?

tyru: sorry.
on second thought, I think it's too early optimization now.
If we do it, let's open a new pull request.

PR #571, opened 2018-01-15
Web.XML: Fix handling empty attribute values

rhysd:
When parsing `<tag attr=""/>`, currently the value of `attr` is parsed as `'attr'`. But it should be `''` in XML. I fixed the point.

ref: https://twitter.com/mattn_jp/status/952766828199030784

### DIFF
--- a/autoload/vital/__vital__/Web/XML.vim
+++ b/autoload/vital/__vital__/Web/XML.vim
@@ -15,6 +15,9 @@ endfunction
 let s:__template = { 'name': '', 'attr': {}, 'child': [] }
 
 function! s:decodeEntityReference(str) abort
+  if a:str ==# ''
+    return a:str
+  endif
   let str = a:str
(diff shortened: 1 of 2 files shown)


mattn:
Thank you. Totally okay to me. :+1:

BTW, I'm thinking original code I've wrote is wrong since it's not possible to recognize which XML is presented <input checked/> or <input checked=""/>. I know it's not an issue of this PR. So I will have to send PR which provide flag to set v:null instead of ''.

mattn:
Could you please add test?

rhysd:
ok, I'll work on it tonight. There is no test for `Web.XML` module yet

rhysd:
@mattn I added some minimal tests to confirm this fix.

mattn:
Cool! 👍

rhysd:
@mattn Could you approve this PR if you feel ok?

mattn:
Thanks! :dog:

PR #581, opened 2018-06-05
CI: update reviewdog and use github-pr-check reporter

haya14busa:
Also, drop droneio configs (which almost anyone cannot manage
except haya14busa...)!

https://medium.com/@haya14busa/reviewdog-github-check-improved-automated-review-experience-58f89e0c95f3


haya14busa:
i'll drop test commit after (or before) got LGTM from reviewers.

scripts/run-lint.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
rhysd: I think this script should check it is run in Travis CI environment because some users may run this script in their local.

```sh
if [[ "$TRAVIS" != "true" ]]; then
  echo "This script is intended to be run on Travis CI" 1>&2
  exit 1
fi
```

rhysd: Or renaming this script to `run-lint-on-ci.sh` may be another option. Users can notice it is for CI by its name.

scripts/run-lint.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -ev
+
+REVIEWDOG_VERSION=0.9.9
+VINT_VERSION=0.3.18
+
+mkdir -p ~/bin/ && export export PATH="~/bin/:$PATH"
```
rhysd: `export export` -> `export`

scripts/run-lint.sh
haya14busa: Makes sense. I included both changes.

scripts/run-lint.sh
haya14busa: Thank you for the catch 😆

PR #582, opened 2018-06-10
Deprecate Lua.Prelude module

ujihisa:
I introduced this ages ago, but I haven't actively maintain it. Nobody besides vital core is using this. I'd like to remove this in the future. Let's deprecate this first, so that people can still use just by renaming, and we can get feedback from them if they exist.


tyru:
:+1:

ujihisa:
oh right I need an github approval from you.

tyru:
@thinca @lambdalisue @mattn @rhysd @haya14busa  what do you think?

tyru:
hmm, you should have waited for others for a while? :thinking_face:
Not only Lua.Prelude, someone may want to say deprecation of vital modules, and so on.
However, I thought you didn't need to hurry.

ujihisa:
True. Good point is that users can still use this module just by renaming, but the renaming process will trigger feedback ;)

PR #587, opened 2018-06-10
Make CI faster

### LINKED ISSUE #534: Remove --rubyinterp　とか
1. `has('ruby')/has('perl')` なコードはない
2. 使ってないのに https://travis-ci.org/vim-jp/vital.vim/jobs/256352309#L1641 みたいに設定不備などで ruby 関連で落ちるの辛い

の二点から `--rubyinterp` とか `--perliterp` とか消したい。

面倒臭すぎて https://github.com/vim-jp/vital.vim/pull/513 では消したけど、一応別枠で確認を取りたい。

rhysd:
I made running CI faster.

Before: https://travis-ci.org/vim-jp/vital.vim/builds/390362502
After: https://travis-ci.org/rhysd/vital.vim/builds/390389523

Total: 5min 47sec -> 4min 2sec (1min 45sec faster)

Optimizations:

- Enabled a binary package (bottle) on macOS worker by removing installation options which prevented it. (4m54s -> 2m58s)
  - macOS worker is slow to start. It was a bottleneck.
  - Removed Lua dependency since Lua was not actually used in CI. (`--with-lua` could be omitted by this change)
  - Link `vim` executable directly (`--with-override-system-vim` could be omitted by this change)
- Separate running linters into another job.
  - Since it took 200s.
  - After optimizing macOS worker, the job which run linters was a bottleneck.

I assigned people as reviewers who would be familiar with
- CI scripts
- macOS

Fix #534

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,13 @@ matrix:
       env:
         - VIM_VERSION=master
         - THEMIS_PROFILE=vim-profile-master.txt
-        - RUN_LINT=true
     - os: osx
       osx_image: xcode8.3
       env:
         - THEMIS_PROFILE=vim-profile-osx.txt
+    - os: linux
+      env:
+        - RUN_LINT=true
 
 addons:
   apt:
@@ -53,16 +55,14 @@ script:
   - uname -a
   - which -a vim
   - vim --cmd version --cmd quit
-  - vim --cmd "try | helptags doc/ | catch | cquit | endtry" --cmd quit
-  # - /tmp/vim-themis/bin/themis --runtimepath /tmp/vimproc --reporter dot
-  - /tmp/vim-themis/bin/themis --runtimepath /tmp/vimproc --exclude ConcurrentProcess --reporter dot
-  - ruby scripts/check-changelog.rb
+  - |
+    if [[ "${RUN_LINT}" != "true" ]]; then
+        /tmp/vim-themis/bin/themis --runtimepath /tmp/vimproc --exclude ConcurrentProcess --reporter dot
+    fi
   - if [[ "${RUN_LINT}" == "true" ]] ; then bash scripts/run-lint-on-ci.sh ; fi
 
 after_success:
-  - covimerage write_coverage $THEMIS_PROFILE
-  - coverage xml
-  - bash <(curl -s https://codecov.io/bash)
+  - bash scripts/coverage.sh
 
 notifications:
   webhooks:
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,6 @@ addons:
       - python-dev
       - python3-dev
       - python3-pip
-      - liblua5.1-0-dev
-      - lua5.1
       - ruby-dev
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,9 @@ addons:
     packages:
       - language-pack-ja
       - vim
-      - libperl-dev
       - python-dev
       - python3-dev
       - python3-pip
-      - ruby-dev
 
 install:
   - rvm reset
--- a/scripts/install-vim.sh
+++ b/scripts/install-vim.sh
@@ -8,14 +8,14 @@ case "${TRAVIS_OS_NAME}" in
 		cd /tmp/vim
 		./configure --prefix="${HOME}/vim" --with-features=huge \
 			--enable-perlinterp --enable-pythoninterp --enable-python3interp \
-			--enable-rubyinterp --enable-luainterp --enable-fail-if-missing
+			--enable-rubyinterp --enable-fail-if-missing
 		make -j2
 		make install
 		;;
 	osx)
 		export HOMEBREW_NO_AUTO_UPDATE=1
 		brew update
-		brew install macvim --with-override-system-vim --with-lua
+		brew install macvim --with-override-system-vim
 		;;
 	*)
 		echo "Unknown value of \${TRAVIS_OS_NAME}: ${TRAVIS_OS_NAME}"
--- a/scripts/install-vim.sh
+++ b/scripts/install-vim.sh
@@ -15,7 +15,10 @@ case "${TRAVIS_OS_NAME}" in
 	osx)
 		export HOMEBREW_NO_AUTO_UPDATE=1
 		brew update
-		brew install macvim --with-override-system-vim
+		brew install macvim
+		# Instead of --with-override-system-vim, manually link the executable because
+		# it prevents MacVim installation with a bottle.
+		ln -s "$(brew --prefix macvim)/bin/mvim" "/usr/local/bin/vim"
 		;;
 	*)
 		echo "Unknown value of \${TRAVIS_OS_NAME}: ${TRAVIS_OS_NAME}"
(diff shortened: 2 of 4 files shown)


ujihisa:
Great work! I'll leave proper review for someone who knows CIs better than me, but I like these changes :+1:

> Removed Lua dependency since Lua was not actually used in CI. (--with-lua could be omitted by this change)

Good catch! :+1:

rhysd:
@ujihisa Thank you. I assigned you since you're more familiar with macOS. From macOS (Homebrew) user point of view, I'd be happy if you check the changes.

ujihisa:
(I used to be a professional MacVim user. Currently I have no MacBooks/iMacs...)

.travis.yml
```diff
@@ -55,16 +55,14 @@ script:
...
+    if [[ "${RUN_LINT}" != "true" ]]; then
+        /tmp/vim-themis/bin/themis --runtimepath /tmp/vimproc --exclude ConcurrentProcess --reporter dot
+    fi
   - if [[ "${RUN_LINT}" == "true" ]] ; then bash scripts/run-lint-on-ci.sh ; fi
 
 after_success:
-  - covimerage write_coverage $THEMIS_PROFILE
-  - coverage xml
-  - bash <(curl -s https://codecov.io/bash)
+  - bash scripts/coverage.sh
```
haya14busa: [optional] i slightly prefer moving this line after themis line.
it doesn't need to be run with RUN_LINT=true, so it's a bit confusing.

scripts/install-vim.sh
```diff
@@ -4,18 +4,22 @@ set -ev
...
 	linux)
+		if [[ "${VIM_VERSION}" == "" ]]; then
+			exit
+		fi
 		git clone --depth 1 --branch "${VIM_VERSION}" https://github.com/vim/vim /tmp/vim
 		cd /tmp/vim
 		./configure --prefix="${HOME}/vim" --with-features=huge \
 			--enable-perlinterp --enable-pythoninterp --enable-python3interp \
-			--enable-rubyinterp --enable-luainterp --enable-fail-if-missing
+			--enable-rubyinterp --enable-fail-if-missing
```
tyru: I'm just curious how it would speed up if `--enable-rubyinterp` was removed,
if it is not too much to ask...

lambdalisue:
Note: It's a bit beyond this PR but I've found that using Docker is quite useful to test on Linux.

https://github.com/lambdalisue/gina.vim/blob/master/.travis.yml#L40-L43

scripts/install-vim.sh
rhysd: I did not consider that. Let me try.

.travis.yml
rhysd: Let me confirm: 'themis line' means the line which runs unit tests with themis.vim, am I correct?

If so, we can't move this line to `script` section because coverage should be taken only if CI was successfully done.
The guard `$RUNLINT != true` is in `coverage.sh`, but separating scripts into another file may make understanding what is done harder. 
Do you think we should write the script directly in `.travis.yml` as follows?

```yaml
- |
    if [[ "${RUN_LINT}" == "" ]]; then
      covimerage write_coverage $THEMIS_PROFILE
      coverage xml
      bash <(curl -s https://codecov.io/bash)
    fi
```

rhysd:
> Note: It's a bit beyond this PR but I've found that using Docker is quite useful to test on Linux.

Definitely. Moving to Docker container might be future work to improve this further.

scripts/install-vim.sh
rhysd: @tyru I tried removing `if_perl` and `if_ruby`, but could not see performance improvements https://travis-ci.org/rhysd/vital.vim/builds/390581618

tyru: @rhysd Thanks!!
Then I think `if_perl`, `if_ruby` should be left.

tyru: hmmmmm, @lambdalisue told me that `if_perl` and `if_ruby` causes build error... (on vim-jp #vital-dev slack)
So now I feel *removing language interfaces as possible* is the right policy then...
Now I remember language interfaces sometimes doesn't follow the latest language's change.

tyru: oops, fixed.
:x: if_perl and if_ruby causes build error
:o: if_perl and if_ruby sometimes causes build error

lambdalisue: I don't think `--enable-luainterp` should be removed while some codes in vital.vim relies on that feature.
https://github.com/vim-jp/vital.vim/search?p=1&q=lua&unscoped_q=lua

I'm very welcome to remove `--enable-rubyinterp` or `--enable-perlinterp` while no codes in vital.vim uses that features and it sometime cause build errors.

scripts/install-vim.sh
```diff
@@ -15,7 +15,10 @@ case "${TRAVIS_OS_NAME}" in
 	osx)
 		export HOMEBREW_NO_AUTO_UPDATE=1
 		brew update
-		brew install macvim --with-override-system-vim
+		brew install macvim
+		# Instead of --with-override-system-vim, manually link the executable because
+		# it prevents MacVim installation with a bottle.
+		ln -s "$(brew --prefix macvim)/bin/mvim" "/usr/local/bin/vim"
```
lambdalisue: How about

```
export THEMIS_VIM="$(brew --prefix macvim)/bin/mvim"
```

instead?

scripts/install-vim.sh
raa0121: I think vital needs `+lua/dyn` for `Text/Sexp`

lambdalisue:
Outdated but

1. I think `--enable-luainterp` should be kept
2. I prefer `export THEMIS_VIM=$(brew ...)` instead

rhysd:
@lambdalisue @raa0121 Tests use only `Text/Sexp.vim` (which has only one test case). Other modules which use Lua don't have any tests. So it's a trade-off. 'Give up one test case for `Text/Sexp.vim`' v.s. 'Making CI faster (about 1.4x faster). I don't have any strong opinion for this, actually.

ujihisa:
Please ignore Text.Sexp. I'm planning to deprecate its lua version, along with https://github.com/vim-jp/vital.vim/pull/582 (Text.Sexp and brainf**k modules depended on it)

tyru:
:+1: for ignore

ujihisa:
:clap: let's merge :clap:

tyru:
Still remaining issue exists https://github.com/vim-jp/vital.vim/pull/587#discussion_r194259022

.travis.yml
haya14busa: Cannot we run like this?
Another option is use `&&` to run coverage scripts.

```
-  if [[ "${RUN_LINT}" != "true" ]]; then
     /tmp/vim-themis/bin/themis --runtimepath /tmp/vimproc --exclude ConcurrentProcess --reporter dot
     bash scripts/coverage.sh
   fi
```

One additional good point to do this is that we can upload coverage scripts when another commands failed.

rhysd: It's also ok. I'll choose it. Thank you for your advice.

rhysd: It seems that the fix for this cause some problem on macOS: https://travis-ci.org/vim-jp/vital.vim/jobs/391106600
I don't have time to investigate so let me skip this for now (since it's just an option).

rhysd:
@tyru @ujihisa @haya14busa @lambdalisue @raa0121 Thank you for your reviews

tyru:
:tada::tada::tada:

Issue #592, opened 2018-06-11
Ensure Data.List functions are nondestructive


tyru:
via https://github.com/vim-jp/vital.vim/pull/511#issuecomment-396419319 (thanks @Milly)

* Remove unnecessary `copy()` from test files
* Add tests to check if functions are nondestructive
* Remove unnecessary `copy()` from document
* Note all functions are nondestructive in the document
  * `Data.List.uniq()`, `uniq_by()` are noted already
  * Is it enough to note all functions are nondestructive in introduction section?

ujihisa:
Since existing Data.List functions are not necessarily nondestructive, I'll clarify destructive/nondestructive for each functions in the documentation, as the first step.

ujihisa:
According to the comment we aren't changing Data.List entirely to non-destructive. Since all the functions with `{list}` argument clearly mention if they are destructive or non-destructive, I close this now.

PR #597, opened 2018-06-20
Data.List: Clarify which destructive/non-destructive explicitly

ujihisa:
https://github.com/vim-jp/vital.vim/issues/592

(One thing I noticed was that `sort_by` is non-destructive while `sort` is destructive. I'd like to leave that decision here but you can think for future. Most of Data.List functions are non-destructive, so it might sound reasonable to change `sort` also non-destructive, but I believe the best is to provide both destructive/non-destructive sort and sort_by, since sorting often take giant lists which effects GC time.)

Handy links for reviewers
* test https://github.com/vim-jp/vital.vim/blob/master/test/Data/List.vimspec
* implementation https://github.com/vim-jp/vital.vim/blob/master/autoload/vital/__vital__/Data/List.vim

### DIFF
--- a/doc/vital/Data/List.txt
+++ b/doc/vital/Data/List.txt
@@ -52,19 +52,27 @@ pop({list})				*Vital.Data.List.pop()*
 	Removes the last element from |List| {list} and returns the element,
 	as if the {list} is a stack.
 
+	Destructive. This modifies {list}.
+
 
 push({list}, {val})			*Vital.Data.List.push()*
 	Appends {val} to the end of |List| {list} and returns the list itself,
 	as if the {list} is a stack.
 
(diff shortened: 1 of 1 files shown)


ujihisa:
age

tyru:
すみません、ちょっとしばらく見れないです。 
誰か他の方レビューお願いします。

ujihisa:
age

aiya000:
sort_byまで確認し、力付きました :dog:
よさそう　:dog2:

kazu9su:
なんの権限も無いですがsort_by以降手動で確認しました。
よさそう 👍

ujihisa:
thinca+
aiya000++
tommy+++

PR #599, opened 2018-07-08
Web.HTTP: Support HEAD method for curl client

pocke:
# Problem

Currently `Web.HTTP.request()` function does not work with `HEAD` HTTP method when client is curl.

## reproduce

```vim
let V = vital#of('vital')
let HTTP = V.import('Web.HTTP')

echo HTTP.request('http://example.com', {"method": 'HEAD', "client": 'curl'}) " Stop this program here!
```

# Cause

curl(1) does not support `-X HEAD`.

```bash
$ curl -X HEAD http://example.com
Warning: Setting custom HTTP method to HEAD with -X/--request may not work the 
Warning: way you want. Consider using -I/--head instead.
```

# Solution

Use `--head` option instead of `-X HEAD`.

### DIFF
--- a/autoload/vital/__vital__/Web/HTTP.vim
+++ b/autoload/vital/__vital__/Web/HTTP.vim
@@ -498,7 +498,12 @@ function! s:clients.curl.request(settings) abort
   if has_key(a:settings, 'gzipDecompress') && a:settings.gzipDecompress
     let command .= ' --compressed '
   endif
-  let command .= ' -L -s -k -X ' . a:settings.method
+  let command .= ' -L -s -k '
+  if a:settings.method ==? 'HEAD'
+    let command .= '--head'
+  else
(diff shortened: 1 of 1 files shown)


pocke:
I'm not sure why the build on Travis CI is failing... :thinking: 
https://travis-ci.org/vim-jp/vital.vim/jobs/401416402#L1596-L1630

Can anyone help me?

k-takata:
> `Vim(python3):/must>not&exist/foo:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses`

This is an issue of Vim with Python 3.7. (vim-jp/issues#1175)

tyru:
Now fixed by #601.
Please merge or rebase master.

pocke:
Rebased and force pushed. Thanks!

PR #608, opened 2018-10-10
[Fix CI] write_coverage -> write-coverage

tyru:


tyru:
* `covimerage` changed subcommand: `write_coverage` -> `write-coverage`
* Default `covimerage` output filename (`--data-file`) is `.coverage_covimerage`, but `coveragerc` says `.coverage.covimerage`
    * So coverage.py tries to find the file but it doesn't exist
    * (why did it work? it may not be changed from before, and `--data-file` is not specified)

tyru:
please approve this...

tyru:
`write_coverage` -> `write-coverage` change was caused by bug, it seems `write_coverage` is the right name.
So we have to revert `write-coverage` to `write_coverage` after the latest one is uploaded to pypi.

.coveragerc
```diff
@@ -1,3 +1,3 @@
 [run]
 plugins = covimerage
-data_file = .coverage.covimerage
+data_file = .coverage_covimerage
```
blueyed: JFI: this is required since the default in covimerage changed (because `.coverage.foo` gets picked up typically with `coverage combine` (for a `.coverage` data_file`) - therefore I've changed it to `.coverage_foo`).  Slightly related: https://github.com/Vimjas/covimerage/issues/51

tyru: got it. thanks for information!

PR #617, opened 2018-11-06
System.File: quote file names in command arguments

### LINKED ISSUE #616: File.copy doesn't work with spaces in the path
https://github.com/previm/previm/issues/9

https://github.com/previm/previm/issues/66

File.copy doesn't work correctly if paths contains spaces.

mattn:
Also fixing encoding of arguments

Fixes #616 

For fixing this, I removed part throwing exceptionin rmdir and fallback to rm-rf command since there are no sense of unity for code. I may add this again.

```vim
    if v:shell_error	
      let ret = iconv(ret, 'char', &encoding)	
      throw 'vital: System.File: rmdir(): ' . substitute(ret, '\n', '', 'g')	
    endif
```

```vim
 if &shell =~? 'sh$'	    " All characters must be encoded to system encoding.
      let cmd = flags =~# 'r' ? 'rm -rf' : 'rmdir'	    if s:need_trans
      let ret = system(cmd . ' ' . shellescape(a:path))	      let path  = iconv(path, &encoding, 'char')
    else
```

I think we should decide what should do here.

* Arguments must be converted to &encoding to char? (yes, fixed)
* Need to quote with double-quote in the command? (yes, fixed)
* Need to throw exception when the command fail?
* Need to do fallback with another command?

Any thought?

### DIFF
--- a/autoload/vital/__vital__/System/File.vim
+++ b/autoload/vital/__vital__/System/File.vim
@@ -117,13 +117,15 @@ elseif s:is_windows
     let dest = substitute(dest, '[/\\]\+', '\', 'g')
     " src must not have trailing '\'.
     let src  = substitute(src, '\\$', '', 'g')
+
     " All characters must be encoded to system encoding.
     if s:need_trans
       let src  = iconv(src, &encoding, 'char')
       let dest = iconv(dest, &encoding, 'char')
     endif
+
     let cmd_exe = (&shell =~? 'cmd\.exe$' ? '' : 'cmd /c ')
-    call system(cmd_exe . 'move /y ' . src  . ' ' . dest)
+    call system(cmd_exe . 'move /y "' . src  . '" "' . dest . '"')
     return !v:shell_error
   endfunction
 else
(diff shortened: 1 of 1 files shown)


autoload/vital/__vital__/System/File.vim
```diff
@@ -312,22 +329,19 @@ elseif s:is_unix
...
+      let path  = iconv(path, &encoding, 'char')
     endif
-  endfunction
 
+    " 'f' flag does not make sense.
+    let cmd = 'rmdir /Q'
+    let cmd .= flags =~# 'r' ? ' /S' : ''
+    call system(cmd . ' "' . path . '"')
+    return !v:shell_error
+  endfunction
```
tyru: In order to keep old behavior, it should throw an exception with command output?

```vim
    let ret = system(cmd . ' "' . path . '"')
    if v:shell_error
      throw 'vital: System.File: rmdir(): ' . substitute(ret, '\n', '', 'g')
    endif
```

autoload/vital/__vital__/System/File.vim
```diff
@@ -117,13 +117,15 @@ elseif s:is_windows
...
+
     " All characters must be encoded to system encoding.
     if s:need_trans
       let src  = iconv(src, &encoding, 'char')
       let dest = iconv(dest, &encoding, 'char')
     endif
+
     let cmd_exe = (&shell =~? 'cmd\.exe$' ? '' : 'cmd /c ')
-    call system(cmd_exe . 'move /y ' . src  . ' ' . dest)
+    call system(cmd_exe . 'move /y "' . src  . '" "' . dest . '"')
```
tyru: Oh, I didn't know that Windows cannot use `"` in filename... 😅

autoload/vital/__vital__/System/File.vim
```diff
@@ -312,22 +329,19 @@ elseif s:is_unix
 
 elseif s:is_windows
   function! s:rmdir(path, ...) abort
+    let path = a:path
     let flags = a:0 ? a:1 : ''
-    if &shell =~? 'sh$'
```
tyru: Yes, I think it's OK to ignore linux shell support in Windows 😓

ujihisa:
ping

PR #632, opened 2019-01-28
Install macvim using "addons" directive

ichizok:


tyru:
ref. https://docs.travis-ci.com/user/reference/osx/#homebrew

ujihisa:
:apple::+1:

PR #634, opened 2019-02-15
Fix error catching

thinca:
Some `:catch`s use incomplete patterns for errors.


thinca:
I am waiting for this to be fixed.
https://github.com/vim/vim/pull/3969

Issue #635, opened 2019-03-09
Web.HTTP Does not support HTTP/2


rbtnn:
https://github.com/rhysd/wandbox-vim/issues/10#issuecomment-471178152

PR #636, opened 2019-03-09
Web.HTTP support HTTP/2

rbtnn:
https://github.com/vim-jp/vital.vim/issues/635

### DIFF
--- a/autoload/vital/__vital__/Web/HTTP.vim
+++ b/autoload/vital/__vital__/Web/HTTP.vim
@@ -191,7 +191,7 @@ function! s:_build_response(header, content) abort
 
   if !empty(a:header)
     let status_line = get(a:header, 0)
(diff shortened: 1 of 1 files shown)


mattn:
I prefer to be strictly. So
```
'^HTTP/[12]\.\d\s\+\(\d\+\)\s\+\
```

rbtnn:
@mattn There are a case where HTTP/2's header does not contain `.`.  
Please see https://github.com/rhysd/wandbox-vim/issues/10#issuecomment-471178152

rbtnn:
It might prefer to be `\(\d\|\d\.\d\)`  than `[0-9.]\+`.

mattn:
I'm thinking illegal value must be error.

HTTP/0.0
HTTP/3
HTTP/999999999999999999

We should include rhysd's patch.

rbtnn:
@mattn  I see. I'll include rhysd's patch.

PR #644, opened 2019-04-27
Proposal: Interface Extension of MD5

tsuyoshicho:
Currently MD5 has only sum (like md5sum).

However, there is a shortage of APIs to use as components.

Proposal:

* sum : return hash string
* digest : return hash bytes list (new, easy to use for other process)

and matrix

* x(str) : accept string data
* x_raw(bytes) : accept bytes list

after PRed this API support.

Thanks.

### DIFF
--- a/autoload/vital/__vital__/Hash/MD5.vim
+++ b/autoload/vital/__vital__/Hash/MD5.vim
@@ -38,14 +38,27 @@ let s:table = [
       \ 0xf7537e82, 0xbd3af235, 0x2ad7d2bb, 0xeb86d391,
       \ ]
 
-
 function! s:sum(data) abort
+  let bytes = s:_str2bytes(a:data)
+  return s:sum_raw(bytes)
+endfunction
+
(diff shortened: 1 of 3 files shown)


tsuyoshicho:
> x(str) : accept string data
> x_raw(bytes) : accept bytes list

maybe type adaptive (type(a:data) is... check)

ujihisa:
@nicwest what do you think of these new 3 APIs? (seeing `Maintainer: nicwest   <oss@nic-west.com>` line)

ujihisa:
TODO: add tests if we are on the right path

tsuyoshicho:
Add testcases and small bug fix

nicwest:
looks good to me

ujihisa:
thx!

https://ci.appveyor.com/project/vim-jp/vital-vim/builds/24175108/job/4wv2vg9c7f4kudn7 tests passed but covimrage failed for some reason. It doesn't seem to be because of this pull request, so it should be safe to merge.

PR #645, opened 2019-04-29
Data.Base64: Add testcase for RFC TestVector

tsuyoshicho:
see #640
and see #641 (mistake closed)

Add RFC TestVector to Data.Base64

There is coverage improvement. (by passing the check pass in padding)

If it's a good one, Please merge it.


ujihisa:
https://ci.appveyor.com/project/vim-jp/vital-vim/builds/24171809tests passed but covimrage failed for some reason. It doesn't seem to be because of this pull request, so it should be safe to merge.

PR #648, opened 2019-05-19
Proposal: Interface Extension of Base64/Base32

tsuyoshicho:
Currently Base64/32 do not have byte-list from/to API.

Therefore, there is a shortage of APIs to use as components.

Proposal:

```
Bytes-list encode to BaseX
encodebytes({bytes})			
	Return baseX encoded string from {bytes}.

Return decoded Bytes-list
decoderaw({str})			
	Return decoded bytes-list from {str} that's baseX encoded.
```

Thanks.

### DIFF
--- a/autoload/vital/__vital__/Data/Base32.vim
+++ b/autoload/vital/__vital__/Data/Base32.vim
@@ -5,19 +5,25 @@ let s:save_cpo = &cpo
 set cpo&vim
 
 function! s:encode(data) abort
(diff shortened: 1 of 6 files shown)


tsuyoshicho:
Note:
1. If you have a conflict with #647, fix one when it gets in the other.
2. There is no additional testing as it is internally verified.

ujihisa:
Thx!

PR #649, opened 2019-05-25
Data.Base32: fix encode/decode 32bit env support

tsuyoshicho:
Problem:
It did not operate in the environment which is not `num64` because the variable manipulation of size more than 4 bytes in Data.Base32

Solution:
Processing was divided and processed within 4 bytes

### DIFF
--- a/autoload/vital/__vital__/Data/Base32.vim
+++ b/autoload/vital/__vital__/Data/Base32.vim
@@ -4,6 +4,15 @@
 let s:save_cpo = &cpo
 set cpo&vim
 
+function! s:_vital_loaded(V) abort
+  let s:V = a:V
+  let s:bitwise = s:V.import('Bitwise')
+endfunction
+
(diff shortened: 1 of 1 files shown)


autoload/vital/__vital__/Data/Base32.vim
```diff
@@ -68,19 +77,53 @@ function! s:_b32decode(b32, table, pad) abort
...
+    "            |------||------||------||------|
+    "  1         -+
+    "  2           +---+
+    "  3                +---+
+    "  4                     +---+
+    "  5                          +---+
+    "  6                               +---+
+    "  7                                    +---+
+    let n_hi = s:bitwise.or(
+          \ s:bitwise.and(s:bitwise.lshift(a2i[a:b32[i]], 3), 0b11111000),
```
tsuyoshicho: パースミス?なのでしょうか?

ujihisa:
thx!

Issue #664, opened 2019-06-29
Web.URI: `s:_eat_scheme` is not defined anywhere


ujihisa:
@tyru 
https://github.com/vim-jp/vital.vim/blob/master/autoload/vital/__vital__/Web/URI.vim#L128 this `s:_eat_scheme()` function is not defined anywhere, so if Vim runtime evaluate it it always throws an error.

vint provided me a meaningful hint.

```
autoload/vital/__vital__/Web/URI.vim:128:24: Undefined variable: s:_eat_scheme (see :help E738)
autoload/vital/__vital__/Web/URI.vim:135:25: Undefined variable: s:_eat_query (see :help E738)
autoload/vital/__vital__/Web/URI.vim:142:28: Undefined variable: s:_eat_fragment (see :help E738)
autoload/vital/__vital__/Web/URI.vim:204:24: Undefined variable: s:_eat_path_abempty (see :help E738)
autoload/vital/__vital__/Web/URI.vim:212:26: Undefined variable: s:_eat_path_absolute (see :help E738)
autoload/vital/__vital__/Web/URI.vim:214:26: Undefined variable: s:_eat_path_noscheme (see :help E738)
autoload/vital/__vital__/Web/URI.vim:216:26: Undefined variable: s:_eat_path_rootless (see :help E738)
autoload/vital/__vital__/Web/URI.vim:238:28: Undefined variable: s:_eat_userinfo (see :help E738)
autoload/vital/__vital__/Web/URI.vim:245:22: Undefined variable: s:_eat_host (see :help E738)
autoload/vital/__vital__/Web/URI.vim:248:24: Undefined variable: s:_eat_port (see :help E738)
autoload/vital/__vital__/Web/URI.vim:431:25: Undefined variable: s:_eat_query (see :help E738)
autoload/vital/__vital__/Web/URI.vim:437:28: Undefined variable: s:_eat_fragment (see :help E738)
autoload/vital/__vital__/Web/URI.vim:476:24: Undefined variable: s:_eat_path_abempty (see :help E738)
autoload/vital/__vital__/Web/URI.vim:483:26: Undefined variable: s:_eat_path_absolute (see :help E738)
autoload/vital/__vital__/Web/URI.vim:485:26: Undefined variable: s:_eat_path_noscheme (see :help E738)
```

tyru:
[...]

PR #666, opened 2019-06-30
Web.URI: Cosmetic change to expand macro

ujihisa:
@tyru 

* Closes https://github.com/vim-jp/vital.vim/issues/664
* This solves all the vint violations for Web.URI
* This also helps debugging, since you can jump to the definition easily.


tyru:
Thanks!

PR #671, opened 2019-06-30
Random: comment for vint

ujihisa:
* Now `docker run --rm -ti -v `pwd`:/src:ro -w /src fnichol/vint --color .` return nothing. Big success!


ujihisa:
:goodpoem:

PR #675, opened 2019-07-03
Proposal: Bitwise enhance that is new method imported from vital-codec library's Vim.Type.Number module

tsuyoshicho:
# Proposal new library

This library operate Number data

- uint8/16/32/64 cast operator (bit mask)
- rotate8/16/32/64 r/l rotate operator

---

Numberについての型的な処理です。

rotate系はbitwiseとどちらか悩ましいですが、固定長の整数型のビット操作、という意味ではどちらもアリか、と思い、こちらに入れてます。

see #667 
ID.UUID
 Data.List.Byte
  Vim.Type.Number
の依存関係から、まずVim.Type.NumberのPRになります。


ujihisa:
Hmm, as you've struggled, these functions are kind of in between Bitwise and Number. In my opinion, these functions must belong to Bitwise instead. I expect "Vim.Type.Number" to be more [Vim's number](https://vimhelp.org/eval.txt.html#expr-number) specific utilities. Such that would not only include truncating upper bits of the internal representation of a number, but also include modulo, factorial, prime numbers, floating point numbers, or other varieties of them.

tyru:
:+1: to move the functions to Bitwise module.

tsuyoshicho:
OK move to Bitwise

autoload/vital/__vital__/Bitwise.vim
```diff
@@ -77,122 +90,207 @@ function! s:_vital_created(module) abort
...
+  else
+    let s:mask32bit = s:or(
+          \ s:lshift(0xFFFF, 16),
+          \          0xFFFF
+          \)
+  endif
+endfunction
+
+" define and/or/xor/invert when no-exist built-in, vim script implementation.
+if !s:builtin_functions_exist
```
rhysd: vital.vim supports Vim 8.0+. And these polyfill functions were introduced at 7.3.377. So these non-builtin functions are not necessary and we can simply remove `if s:builtin_functions_exist | finish | endif` check IMO.

tsuyoshicho: Thanks great!

Re-tune code and remove Vim.Type.Number, coming soon.

tsuyoshicho: Fixed
Then rebase code treat fixup and Vim.Type.Number drop from Git history.
But keep this review info.

tsuyoshicho:
Hmmm.

Vint review :
- :ng: vint check skip
- :ok: replace built-in

stay tune

PR #676, opened 2019-07-04
Random.shuffle(): mattn suggestion implementation

tsuyoshicho:
see https://vim-jp.slack.com/archives/C8F0X4DHN/p1562239287236300

This implementation is technical trial.

Please review it.

### DIFF
--- a/autoload/vital/__vital__/Random.vim
+++ b/autoload/vital/__vital__/Random.vim
@@ -68,7 +68,7 @@ endfunction
 function! s:Random.shuffle(list) abort
   let pos = len(a:list)
   while 1 < pos
-    let n = self.range(pos-1)
+    let n = self.range(pos)
     let pos -= 1
     let [a:list[n], a:list[pos]] = [a:list[pos], a:list[n]]
   endwhile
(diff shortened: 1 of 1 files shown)


tsuyoshicho:
#677 happen.

autoload/vital/__vital__/Random.vim
```diff
@@ -70,16 +70,11 @@ function! s:Random.shuffle(list) abort
   while 1 < pos
     let n = self.range(pos)
     let pos -= 1
-    if n != pos
-      let temp = a:list[n]
-      let a:list[n] = a:list[pos]
-      let a:list[pos] = temp
-    endif
+    let [a:list[n], a:list[pos]] = [a:list[pos], a:list[n]]
```
thinca: Using list assignment to swap is OK. 👌
I have doubts about removing `if n != pos`. 🤔
There is needless swapping when it is removed.

tsuyoshicho: Hmmm.

Okay, re-fix for later.

tsuyoshicho: fixit

PR #683, opened 2019-07-23
Proposal: Vim.Type: type map/reverse map auto generate

tsuyoshicho:
## Problem:
Vim.Type have type map / type reverse map.
But it is maintained at manually.

## Solution:
Vim.Type type map / type reverse map generated from one typename list.
Maintenance simply.

### DIFF
--- a/autoload/vital/__vital__/Vim/Type.vim
+++ b/autoload/vital/__vital__/Vim/Type.vim
@@ -1,31 +1,16 @@
-let s:types = {
-\   'number'  : 0,
-\   'string'  : 1,
-\   'func'    : 2,
-\   'list'    : 3,
-\   'dict'    : 4,
-\   'float'   : 5,
-\   'bool'    : 6,
-\   'none'    : 7,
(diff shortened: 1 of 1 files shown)


autoload/vital/__vital__/Vim/Type.vim
```diff
@@ -1,31 +1,16 @@
...
-\   'channel' : 9,
-\   'blob'    : 10,
-\ }
-lockvar 1 s:types
+let s:typelist = ['number', 'string', 'func', 'list', 'dict', 'float', 'bool', 'none', 'job', 'channel', 'blob']
+
+let s:types = {}
+let s:type_names = {}
+
+for i in range(len(s:typelist))
```
ujihisa: Is this `i` global?

tsuyoshicho: Ou...

Renew to script-local.
Just a moment.

ujihisa:
:goodpoem:

PR #690, opened 2019-09-11
Proposal : Data.List.Byte as Byte-like data operator and converter

tsuyoshicho:
Byte-like List (0-255 number list) convert from/to other type

(It need  #675 then include commit. When merge timing, rebase it)


doc/vital/Data/List/Byte.txt
```diff
@@ -0,0 +1,70 @@
...
+*Vital.Data.List.Byte* is bytes |List| and other bytes-list like data converter.
+
+==============================================================================
+INTERFACE				*Vital.Data.List.Byte-interface*
+------------------------------------------------------------------------------
+FUNCTIONS				*Vital.Data.List.Byte-functions*
+
+validate({data})		         *Vital.Data.List.Byte.validate()*
+	Return bool value.
+	{data} is List, all data as |Number| and it is min 0/max 255,
```
ujihisa:
```suggestion
	returns 1 the given {data} is a |List| with its all values are 0 to 255 |Number|s.
```

ujihisa: urg it doesn't support multiline fixes

ujihisa: Added 2bdfb5a9272a738c7b5dcc1b7c20686bee93a9c4 a6ad36dcb64ad417dd67b71b4a4daff144458527 at master instead. These are trivial doc fixes and don't need to be checked by CIs.

ujihisa:
Great thx!

doc/vital/Data/List/Byte.txt
tsuyoshicho: Thanks a lot!!

PR #693, opened 2019-09-24
Prelude: Remove escape_file_searching()

ujihisa:
* This has been undocumented
* This function doesn't seem to be actively directly used from Vim plugins,
  but only used from within Prelude
* I'd like to shrink Prelude module gradually, so let's deprecate this one for now.


ujihisa:
ping

PR #695, opened 2019-11-11
Web.json v8

Milly:
`Web.JSON` is too old.
Update to vim 8.

...But since `json_encode()` already exists, is this module unnecessary?

This PR includes these new featues.

* 'allow_nan' setting.
* 'ensure_ascii' setting.
* 'from_encoding' setting.


lambdalisue:
From my point of view, *at least this module* should keep backward compatibility while

1. **People should use `json_encode()` instead as you said**
2. That's why there is no positive reason to improve this module
3. Code which uses this module uses this module may because of some reason

Milly:
> 1. **People should use `json_encode()` instead as you said**

Vital.vim is currently (basically) support vim 8.0 or later.
So should this module be deprecated?
Write to document "deprecated" and "should use json_encode()"?

> 2. That's why there is no positive reason to improve this module

I'm thinking of an implementation that makes the JSON module more flexible like [Python's JSON Library](https://docs.python.org/3.6/library/json.html).
`json_encode()` is [not-strict](https://github.com/vim-jp/issues/issues/1302) and not-flexible.
Should this be a separate module? (eg. **Data**.JSON or ...)

> 3. Code which uses this module uses this module may because of some reason

Agree

One thing, These are purely bug fixes.
* Convert control chars. (Encoded invalid JSON)
* Convert UTF-16 surrogate pairs. (Can not decode valid JSON)

lambdalisue:
> So should this module be deprecated?
> Write to document "deprecated" and "should use json_encode()"?

Yeah, I think so but it's just **my** opinion.

> I'm thinking of an implementation that makes the JSON module more flexible like Python's JSON Library.
> json_encode() is not-strict and not-flexible.
> Should this be a separate module?

Got it. Well, then I think it's better to

1. Use different namespace as you said (I'm OK with `Data.JSON`)
2. Remove any codes for backward compatibilities
3. Apply bug fixes to `Web.JSON` (Optional, if you desire)

autoload/vital/__vital__/Web/JSON.vim
```diff
@@ -42,78 +68,110 @@ endfunction
...
+  if n < 0x10000
+    return printf('\u%04x', n)
+  else
+    let n -= 0x10000
+    return printf('\u%04x%\u%04x', 0xd800 + n / 0x400, 0xdc00 + and(0x3ff, n))
+  endif
+endfunction
+
+function! s:_encode(val, settings) abort
+  let t = type(a:val)
```
tsuyoshicho, line 134: If it doesn't conflict with your change policy, change to comparing it with `Vim.Type` ?

doc/vital/Web/JSON.txt
```diff
@@ -65,21 +70,45 @@ encode({object}[, {settings}])		*Vital.Web.JSON.encode()*
...
 decode({json}[, {settings}])		*Vital.Web.JSON.decode()*
 	Decode a JSON string into an object that vim can treat.
 	{settings} is a |Dictionary| which allows the following:
 
-	'use_token'
-	Use special tokens (e.g. |Vital.Web.JSON.true|) to represent
-	corresponding javascript tokens (e.g. 'true').
-	Otherwise 1 or 0 are used to represent these.
-	The default value is 0.
+        'allow_nan'
```
tsuyoshicho, line 102: This point is incompatible changing. need appending item to `Changes`.

Please `Changes` appending to commit after all changes are done, and before PR merging.

Milly: Is the point you say **Add 'allow_nan'** or **Remove 'use_token'**?

autoload/vital/__vital__/Web/JSON.vim
Milly: Changed it.

doc/vital/Web/JSON.txt
tsuyoshicho: Yes.

I think, Adding phrase to `Changes` for example `Improve and *use_token* remove, and option renewal to  support *allow_nan*` .

Milly:
I'm created #743 that part of bug fixes of this PR.

mattn:
Closed by https://github.com/vim-jp/vital.vim/pull/743 ?

Milly:
> Closed by #743 ?

It's includes only bug fixes that is part of this.

I'm rebased and force pushed.

PR #696, opened 2019-11-19
Proposal: update and simply refine Hash(MD5,SHA1) module; using Data.List.Byte functions

tsuyoshicho:
Update Hash.MD5,Hash.SHA1.

Using previous PRed Data.List.Byte. 

And small add more.
module has Name(using crypto suite eg.HMAC-MD5, HMAC-SHA1) and length


tsuyoshicho:
Thanks for speedy merge!

PR #705, opened 2020-01-07
Minor fix.

LumaKernel:


ujihisa:
CIs kept failing at random place. I'll skip and force merging this.

PR #706, opened 2020-01-17
fix: Data.String: nsplit work same as split()

### LINKED ISSUE #704: Vital.Data.String.nsplit() doesn't keep empty
I expect `keepempty` flag keeps empty strings in the list, but they are not kept.

i.e.

```vim
let s:V = vital#foo#new()
call s:V.load('Data.String')

echom json_encode(split('bar::', ':', 1))
echom json_encode(s:V.Data.String.nsplit('bar::', 10, ':', 1))
```

->

```
['bar', '', '']
['bar']
```

tsuyoshicho:
close #704

Refine `nsplit()` implement like `split()`.


tsuyoshicho:
Add test later

tsuyoshicho:
Mac test fails other reasons.

ujihisa:
Cont'd to https://github.com/vim-jp/vital.vim/pull/708 for MacVim

tsuyoshicho:
I'm OK.

PR #710, opened 2020-01-19
Add check v:version

kazukazuinaina:
Hi!
In this `README.md`, vital.vim support ` Vim 8.1 or later`.
Therefore, I added checking about v:version < 800 in plugin dir.
Could you check this Pull Request?

### DIFF
--- a/plugin/vitalizer.vim
+++ b/plugin/vitalizer.vim
@@ -2,6 +2,8 @@
 
 if exists('g:loaded_vitalizer')
   finish
+elseif v:version < 800
+  echoerr "vital.vim does not work this version of Vim"
 endif
 let g:loaded_vitalizer = 1
 


tsuyoshicho:
Thanks, I suggest to slack vital-dev channel.

kazukazuinaina:
Ok, I got it.

kazukazuinaina:
@ujihisa 
Thanks for review.
I added a error message in this patch.

ujihisa:
I'd also like to hear @thinca, @lambdalisue, @aiya000, @haya14busa, and @tyru's thoughts (I understand @tyru is super busy, so mostly from other people) if we should let Vim fail fast or leave it running.

thinca:
I can accept with squash.

kazukazuinaina:
@thinca 
I squashed.

plugin/vitalizer.vim
```diff
@@ -2,6 +2,8 @@
 
 if exists('g:loaded_vitalizer')
   finish
+elseif v:version < 800
```
thinca: This accepts Vim 8.0.  Is this intentional?

lambdalisue:
I totally agree with @ujihisa I don't think vital itself should check the version.

plugin/vitalizer.vim
kazukazuinaina: Yes, I wrote this patch intentionally to exclude libraries that don't work with Vim7.

kazukazuinaina:
If you don't need this pull request, feel free to close it. 🙇

lambdalisue:
I'm going to close this tomorrow unless someone wants this feature.

haya14busa:
1. Description of version restriction in README is not for vitalizer which is only used by plugin developers, but for vital modules that are used by plugin users. The change itself is wrong if you want to add version check to vital as modules according to README.

2. I also think we don't need to check version in vital "modules" as a whole. Readme says it "guarantee" the following version and not strict requirements too. 

3. If you mean we want to add a version check to "vitializer", it's fine but I'm not sure it worth it.
If you or someone want to use Vim 8+ feature in vitalizer, please go ahead.

tyru:
I *totally* agree with @haya14busa

kazukazuinaina:
Ok, I close this Pull Request.
Thanks for checking this Pull Request.

Issue #698, opened 2019-11-24
Test Fail: MacVim latest (2019 fall) Async.Promise resolve test fails


tsuyoshicho:
MacVim tests fail recently (as of 2019-11) as

```log
Failures:
  1) Async.Promise .resolve() should create resolved Promise with given Promise value
     Error occurred line:
       2:         Assert Equals(p._state, FULFILLED)
     The equivalent values were expected, but it was not the case.
     
         expected: 1
              got: 0
```

e.g.

https://travis-ci.org/vim-jp/vital.vim/jobs/611141864

I guess I should issue at MacVim if it's MacVim specific. It seems like it's a timing problem since it sometimes succeeds.

These are the targeted tests. Please check these if you have MacVim. I don't have MacVim env locally.

```vim
" need vim-jp/vital.vim install

function! s:wait_has_key(obj, name) abort
  let i = 0
  while i < 500
    sleep 10m
    if has_key(a:obj, a:name)
      return
    endif
    let i += 1
  endwhile
  throw printf("s:wait_has_key(): After 5000ms, the given object does not have key '%s': %s", a:name, a:obj)
endfunction

function s:testsuite() abort
  let P = vital#vital#import('Async.Promise')

  " Utilities
  let Wait = {ms -> P.new({res -> timer_start(ms, res)})}
  for p in [
        \   P.resolve(P.resolve(42)),
        \   P.resolve(P.resolve(P.resolve(P.resolve(42)))),
        \   P.resolve(Wait(10).then({-> 42}))
        \ ]
    let l = l:
    call p.then({x -> extend(l, {'result' : x})})
    call s:wait_has_key(l, 'result')
    echo result
    unlet result
  endfor
endfunction

call s:testsuite()
```

Note that these succeed quickly in my non-MacVim Vim.

ujihisa:
https://github.com/vim-jp/vital.vim/pull/707

ujihisa:
@tsuyoshicho translated your messages for the ease of access.

tsuyoshicho:
Thanks

tsuyoshicho:
Fix this issue in #713

PR #713, opened 2020-01-20
Fix Async.Promise tests for test stabiligy even with MacVim

ujihisa:
https://github.com/vim-jp/vital.vim/issues/698

### DIFF
--- a/test/Async/Promise.vimspec
+++ b/test/Async/Promise.vimspec
@@ -11,7 +11,7 @@ function! s:wait_has_key_with_timeout(obj, name, timeout_sec) abort
     endif
     let i += 1
   endwhile
-  throw printf("s:wait_has_key(): After 5000ms, the given object does not have key '%s': %s", a:name, a:obj)
+  throw printf("s:wait_has_key(): After %ds, the given object does not have key '%s': %s", a:timeout_sec, a:name, a:obj)
 endfunction
 
 function! s:resolver(resolve, reject) abort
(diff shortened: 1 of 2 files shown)


tsuyoshicho:
Reviewed-by: Tsuyoshi CHO

(Then wait for lambdalisue's reviewing result or just merge now?)

test/Async/Promise.vimspec
```diff
@@ -606,10 +610,10 @@ Describe Async.Promise
       \   P.resolve(P.resolve(P.resolve(P.resolve(42)))),
       \   P.resolve(Wait(10).then({-> 42}))
       \ ]
-        Assert Equals(p._state, FULFILLED)
         let l = l:
         call p.then({x -> extend(l, {'result' : x})})
         call s:wait_has_key(l, 'result')
+        Assert Equals(p._state, FULFILLED)
```
lambdalisue: 📚 

> The Promise.resolve() method returns a Promise object that is resolved with a given value. If the value is a promise, that promise is returned; if the value is a thenable (i.e. has a "then" method), the returned promise will "follow" that thenable, adopting its eventual state; otherwise the returned promise will be fulfilled with the value. This function flattens nested layers of promise-like objects (e.g. a promise that resolves to a promise that resolves to something) into a single layer.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

Thus `p._state` may not be fulfilled just after `resolve()` if the value is promise or thenable.
That's why the assert timing has changed.

Nice catch 🎉

ujihisa:
It depends; in this case it's not user-facing changes since the changes are only within test directory. I think it's safer to wait for other people's opinions too only if the changes impact the users or the architecture of vital.vim itself.

test/Async/Promise.vimspec
```diff
@@ -1,13 +1,17 @@
 function! s:wait_has_key(obj, name) abort
+  return s:wait_has_key_with_timeout(a:obj, a:name, 5)
+endfunction
+
+function! s:wait_has_key_with_timeout(obj, name, timeout_sec) abort
```
lambdalisue: may

I feel it's better to remove `s:wait_has_key_with_timeout` and make a single function `s:wait_has_key` because

1. Reducing the number of function helps to reduce the number of noise
2. The error message said `s:wait_has_key()` so it's a bit confusing

test/Async/Promise.vimspec
ujihisa: Yay!

test/Async/Promise.vimspec
ujihisa: Makes sense, thanks! How's this? https://github.com/vim-jp/vital.vim/pull/714

PR #716, opened 2020-02-03
Improve performance of Promise especially in Windows

lambdalisue:
I've relied on `Async.Promise` to make [fern.vim](https://github.com/lambdalisue/fern.vim) but the performance issue occurred in Windows.

https://github.com/lambdalisue/fern.vim/pull/57

This PR introduces an event loop and workers to execute tasks. Approx. 20x performance improvement has observed in above plugin.

# Regression tests

```vim
let s:Promise = vital#vital#import('Async.Promise')

function! s:tick() abort
  let promise = s:Promise.resolve(reltime())
  for i in range(1000)
    let promise = promise.then({ s -> s })
  endfor
  return promise.then({ s -> reltimestr(reltime(s)) })
endfunction

function! s:test() abort
  new
  let bufnr = bufnr('%')
  let promise = s:Promise.all([
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \ s:tick(),
        \])
  call promise.then({ vs -> setbufline(bufnr, 1, map(vs, { _, v -> v })) })
endfunction

call s:test()
```

## Summary

- Slight performance regressions has observed in Neovim on macOS (approx 1.1 -> 1.9)
- Slight performance improveoment has observed in Vim on macOS (approx 2.3 -> 1.7)
- Significant performance improveoment has observed in Vim on Windows (approx. 18.1 -> 3.9)

## Before
```
Neovim v0.4 macOS
1.357061
1.316463
1.264815
1.229064
1.192373
1.154427
1.119024
1.082092
1.041670
0.994399

Vim 8.1 macOS
2.510612
2.465745
2.425529
2.385125
2.342392
2.288333
2.246469
2.204589
2.163871
2.123382

Vim 8.2 Windows
18.334129
18.291351
18.250988
18.196796
18.157503
18.116416
18.075665
18.031709
17.991818
17.951131
```

## After

```
Neovim v0.4 macOS
2.130478
2.092259
2.058482
2.023667
1.990111
1.939237
1.903443
1.868484
1.835093
1.795646

Vim 8.1 macOS
1.939646
1.891795
1.850817
1.809631
1.768790
1.729066
1.687586
1.643133
1.596527
1.554153

Vim 8.2 Windows
4.146574
4.101837
4.060613
4.021967
3.984135
3.946313
3.909356
3.862812
3.818288
3.776003
```

### DIFF
--- a/autoload/vital/__vital__/Async/Later.vim
+++ b/autoload/vital/__vital__/Async/Later.vim
@@ -1,6 +1,7 @@
 let s:tasks = []
 let s:workers = []
 let s:max_workers = 50
+let s:error_handler = v:null
 
 function! s:call(fn, ...) abort
   call add(s:tasks, [a:fn, a:000])
@@ -28,7 +29,7 @@ function! s:set_error_handler(handler) abort
   let s:error_handler = a:handler
 endfunction
 
-function! s:default_error_handler() abort
+function! s:_default_error_handler() abort
   let ms = split(v:exception . "\n" . v:throwpoint, '\n')
   echohl ErrorMsg
   for m in ms
@@ -51,7 +52,11 @@ function! s:_worker(...) abort
   try
     call call('call', remove(s:tasks, 0))
   catch
-    call s:error_handler()
+    if s:error_handler is# v:null
+      call s:_default_error_handler()
+    else
+      call s:error_handler()
+    endif
   endtry
   if n_workers < s:max_workers
     call add(s:workers, timer_start(0, s:Worker, { 'repeat': -1 }))
@@ -59,4 +64,3 @@ function! s:_worker(...) abort
 endfunction
 
 let s:Worker = funcref('s:_worker')
-let s:error_handler = funcref('s:default_error_handler')
--- a/autoload/vital/__vital__/Async/Later.vim
+++ b/autoload/vital/__vital__/Async/Later.vim
@@ -1,21 +1,31 @@
 let s:tasks = []
-let s:timer = v:null
+let s:workers = []
+let s:max_workers = 50
 
-function! s:call(fn, ...) abort
+function! s:call(fn, ...) abort dict
   call add(s:tasks, [a:fn, a:000])
-  call s:_emit()
+  if empty(s:workers)
+    call add(s:workers, timer_start(0, s:Worker, { 'repeat': -1 }))
+  endif
 endfunction
 
-function! s:_emit() abort
-  if v:dying || s:timer isnot# v:null || empty(s:tasks)
+function! s:get_max_workers(n) abort
+  return s:max_workers
+endfunction
+
+function! s:set_max_workers(n) abort
+  let s:max_workers = a:n
+endfunction
+
+function! s:_worker(...) abort
+  if v:dying
     return
   endif
-  let s:timer = timer_start(0, funcref('s:_callback'))
-endfunction
-
-function! s:_callback(timer) abort
-  let s:timer = v:null
-  if v:dying || empty(s:tasks)
+  let n_workers = len(s:workers)
+  if empty(s:tasks)
+    if n_workers
+      call timer_stop(remove(s:workers, 0))
+    endif
     return
   endif
   try
@@ -28,5 +38,9 @@ function! s:_callback(timer) abort
     endfor
     echohl None
   endtry
-  call s:_emit()
+  if n_workers < s:max_workers
+    call add(s:workers, timer_start(0, s:Worker, { 'repeat': -1 }))
+  endif
 endfunction
+
+let s:Worker = funcref('s:_worker')
--- a/autoload/vital/__vital__/Async/Later.vim
+++ b/autoload/vital/__vital__/Async/Later.vim
@@ -9,7 +9,7 @@ function! s:call(fn, ...) abort
   endif
 endfunction
 
-function! s:get_max_workers(n) abort
+function! s:get_max_workers() abort
   return s:max_workers
 endfunction
 
--- a/autoload/vital/__vital__/Async/Later.vim
+++ b/autoload/vital/__vital__/Async/Later.vim
@@ -1,0 +1,32 @@
+let s:tasks = []
+let s:timer = v:null
+
+function! s:call(fn, ...) abort
+  call add(s:tasks, [a:fn, a:000])
+  call s:_emit()
+endfunction
+
+function! s:_emit() abort
+  if v:dying || s:timer isnot# v:null || empty(s:tasks)
+    return
+  endif
+  let s:timer = timer_start(0, funcref('s:_callback'))
+endfunction
+
+function! s:_callback(timer) abort
+  let s:timer = v:null
+  if v:dying || empty(s:tasks)
+    return
+  endif
+  try
+    call call('call', remove(s:tasks, 0))
+  catch
+    let ms = split(v:exception . "\n" . v:throwpoint, '\n')
+    echohl ErrorMsg
+    for m in ms
+      echomsg m
+    endfor
+    echohl None
+  endtry
+  call s:_emit()
+endfunction
--- a/autoload/vital/__vital__/Async/Later.vim
+++ b/autoload/vital/__vital__/Async/Later.vim
@@ -2,7 +2,7 @@ let s:tasks = []
 let s:workers = []
 let s:max_workers = 50
 
-function! s:call(fn, ...) abort dict
+function! s:call(fn, ...) abort
   call add(s:tasks, [a:fn, a:000])
   if empty(s:workers)
     call add(s:workers, timer_start(0, s:Worker, { 'repeat': -1 }))
--- a/autoload/vital/__vital__/Async/Later.vim
+++ b/autoload/vital/__vital__/Async/Later.vim
@@ -14,6 +14,9 @@ function! s:get_max_workers(n) abort
 endfunction
 
 function! s:set_max_workers(n) abort
+  if a:n <= 0
+    throw 'vital: Async.Later: the n must be a positive integer'
+  endif
   let s:max_workers = a:n
 endfunction
 
--- a/autoload/vital/__vital__/Async/Later.vim
+++ b/autoload/vital/__vital__/Async/Later.vim
@@ -20,6 +20,23 @@ function! s:set_max_workers(n) abort
   let s:max_workers = a:n
 endfunction
 
+function! s:get_error_handler() abort
+  return s:error_handler
+endfunction
+
+function! s:set_error_handler(handler) abort
+  let s:error_handler = a:handler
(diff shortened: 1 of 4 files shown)


lambdalisue:
It seems test become fragile. I'll try to fix

rhysd:
Thank you for adding this.

> approx. 20x in Windows for the plugin above.

Would you also confirm that performance regressions don't happen on other platform?

lambdalisue:
Done. Please review it @rhysd @thinca @mattn 🙇

rhysd:
Let me check this tomorrow or day after tomorrow. I actually need to make another pull request and need to check another issue today...

autoload/vital/__vital__/Async/Later.vim
```diff
@@ -0,0 +1,46 @@
...
+  endif
+  try
+    call call('call', remove(s:tasks, 0))
+  catch
+    let ms = split(v:exception . "\n" . v:throwpoint, '\n')
+    echohl ErrorMsg
+    for m in ms
+      echomsg m
+    endfor
+    echohl None
```
rhysd: I think library should not echo errors. Would you consider how to handle errors? I think adding error handling function to `s:tasks` might be an option.

```vim
function! s:call(errfn, fn, ...) abort
  call add(s:tasks, [a:fn, a:000, errfn])
  ...
endfunction
```

autoload/vital/__vital__/Async/Later.vim
```diff
@@ -0,0 +1,46 @@
+let s:tasks = []
+let s:workers = []
+let s:max_workers = 50
```
rhysd: Just a question, but how did you choose this value?

autoload/vital/__vital__/Async/Later.vim
```diff
@@ -0,0 +1,46 @@
...
+    call add(s:workers, timer_start(0, s:Worker, { 'repeat': -1 }))
+  endif
+endfunction
+
+function! s:get_max_workers(n) abort
+  return s:max_workers
+endfunction
+
+function! s:set_max_workers(n) abort
+  let s:max_workers = a:n
```
rhysd: Please check `a:n > 0`

test/Async/Later.vimspec
```diff
@@ -0,0 +1,77 @@
...
+        let exception = v:exception
+      finally
+        redir END
+      endtry
+      Assert Equals(exception, v:null)
+      Assert Match(
+            \ messages,
+            \ '^\nHello World\nfunction .*<SNR>\d\+_throw, line 1',
+            \)
+    End
```
rhysd: Can you add a test case where number of tasks is greater than number of workers by setting small number to `max_workers`?

autoload/vital/__vital__/Async/Later.vim
```diff
@@ -0,0 +1,46 @@
...
+let s:max_workers = 50
+
+function! s:call(fn, ...) abort
+  call add(s:tasks, [a:fn, a:000])
+  if empty(s:workers)
+    call add(s:workers, timer_start(0, s:Worker, { 'repeat': -1 }))
+  endif
+endfunction
+
+function! s:get_max_workers(n) abort
```
rhysd: `n` is unnecessary

test/Async/Later.vimspec
```diff
@@ -0,0 +1,77 @@
+Describe Async.Later
+  Before all
+    let Later = vital#vital#import('Async.Later')
+    " NOTE: To start event-loop, call void function here
+    call Later.call({ -> 0 })
```
mattn: Would you make this more cool?

test/Async/Later.vimspec
lambdalisue: Sure 👍

autoload/vital/__vital__/Async/Later.vim
lambdalisue: Kind "kan". I tried 10, 30, 50, 100 and 50 was enough to get maximum performance by 2-3 trials.

autoload/vital/__vital__/Async/Later.vim
lambdalisue: Well, I agree that the library should not echo errors in the general case. However I believe that the error handling you mentioned is OK in this case because it is a safety-net which developers should NOT rely on. But I understand that sometime these kind of safety-net is quite annoying thus I added `set_error_handler()` method to replace **entire** error handling instead.

test/Async/Later.vimspec
lambdalisue: Ah, I guess it's no longer necessary. I removed it

autoload/vital/__vital__/Async/Later.vim
lambdalisue: Additionally, I tried to follow APIs of native `call()` function so I don't want to add `errfn` or whatever to `Later.call()`.

rhysd: OK, thank you for the explanation. Sounds it makes sense. I think we should add note that callback of the `Later.call` should handle exception by itself and the error handler is just a safety net.

autoload/vital/__vital__/Async/Later.vim
rhysd: I see. I understood that it was determined based on your use case.

doc/vital/Async/Later.txt
```diff
@@ -0,0 +1,94 @@
...
+FUNCTION 				*Vital.Async.Later-function*
+
+				*Vital.Async.Later.call()*
+.call({task} [, {arglist} [, {dict}]])
+	Similar to |call()| but it calls {task} a bit later, when Vim is not
+	busy. For convenience, {arglist} can be omit.
+
+	When an exception is thrown in {task}, the exception is caught and
+	echomsg with |hl-ErrorMsg| by a default error handler.
+	Use |Vital.Async.Later.set_error_handler()| to change this behavior.
```
rhysd: In favor of https://github.com/vim-jp/vital.vim/pull/716#discussion_r375016263, would you add note that `{task}` should handle an exception by itself and the error handler function is just a safety net?

autoload/vital/__vital__/Async/Later.vim
```diff
@@ -0,0 +1,62 @@
...
+  if empty(s:tasks)
+    if n_workers
+      call timer_stop(remove(s:workers, 0))
+    endif
+    return
+  endif
+  try
+    call call('call', remove(s:tasks, 0))
+  catch
+    call s:error_handler()
```
mattn: If nulll is set via set_error_handler, v:exception should be raised here.

lambdalisue: While the error handling is a safety-net, I've changed behavior to use `s:_default_error_handler()` instead when `s:error_handler` is `v:null`.

lambdalisue:
```
# Run reviewdog.

reviewdog -reporter=github-pr-check

2020/02/05 04:26:23 [vint] reported: https://github.com/vim-jp/vital.vim/runs/426832132

2020/02/05 04:26:23 [vital-throw-message] reported: https://github.com/vim-jp/vital.vim/runs/426832133

reviewdog: Check request failed: Post https://reviewdog.app/check: x509: certificate is valid for invalid.invalid, not reviewdog.app

The command "if [[ "${RUN_LINT}" == "true" ]] ; then bash scripts/run-lint-on-ci.sh ; fi" exited with 1.
```

It seems reviewdog is not available for now? Anyway ignore the CI failure.

lambdalisue:
I'll merge it once @mattn accept it.

rhysd:
:tada:

doc/vital/Async/Later.txt
```diff
@@ -0,0 +1,94 @@
...
+	Note that this error handler exists as a safety-net. Developers should
+	NOT rely on the behavior of the error handler and SHOULD catch errors
+	by themselves in each {task}.
+
+	Internally, it enqueues a {task} into an internal task queue. Tasks
+	in the queue will be dequeued and processed by internal workers. The
+	number of internal workers are gradually increase until all tasks in
+	the queue are completed or reached to the value of "max_workers".
+	Once the queue become empty, the number of internal workers are
+	gradually decrease so that no workers exists at the end.
```
thinca: What is a reason the workers gradually increase/decrease instead of immediate?
Did you encounter performance issues?

test/Async/Later.vimspec
```diff
@@ -0,0 +1,113 @@
...
+        sleep 10m
+      catch
+        let exception = v:exception
+      finally
+        redir END
+      endtry
+      Assert Equals(exception, v:null)
+      Assert Match(
+            \ messages,
+            \ '^\nHello World\nfunction .*<SNR>\d\+_throw, line 1',
```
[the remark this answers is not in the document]
lambdalisue: How should I fix this? IMO, we should enforce English during tests like https://github.com/lambdalisue/fern.vim/blob/master/test/.themisrc#L12-L17

autoload/vital/__vital__/Async/Later.vim
```diff
@@ -0,0 +1,66 @@
...
+  endif
+endfunction
+
+function! s:get_max_workers() abort
+  return s:max_workers
+endfunction
+
+function! s:set_max_workers(n) abort
+  if a:n <= 0
+    throw 'vital: Async.Later: the n must be a positive integer'
```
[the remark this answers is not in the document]
lambdalisue: A term `n` is used in help as well so I think it's ok to use tern `n` here.

doc/vital/Async/Later.txt
lambdalisue: Yes. The first version starts 50 timers at the beginning but sometimes it took more than 10ms in Windows (tests has randomly failed.)

test/Async/Later.vimspec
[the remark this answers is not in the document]
thinca: I agree with fixing message locale 😃

autoload/vital/__vital__/Async/Later.vim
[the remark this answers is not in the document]
thinca: The message will be shown for a user, so it is better if the message is articulate.
But I do not request to fix this strongly.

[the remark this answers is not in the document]
lambdalisue: > The message will be shown for a user,

I think users don't understand anyway even `n` was `max_workers`... So I'll keep it at least for now.

test/Async/Later.vimspec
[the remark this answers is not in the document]
lambdalisue: https://github.com/vim-jp/vital.vim/issues/720

PR #721, opened 2020-02-13
Add Promise.on_unhandled_rejection

hrsh7th:
This PR aims to able to catch all unhandled rejection on Async.Promise.

It will be useful for logging plugin's errors, I think.

### DIFF
--- a/autoload/vital/__vital__/Async/Promise.vim
+++ b/autoload/vital/__vital__/Async/Promise.vim
@@ -38,6 +38,7 @@ let s:NOOP = funcref('s:noop')
 
 let s:PROMISE = {
     \   '_state': s:PENDING,
+    \   '_has_floating_child': v:false,
     \   '_children': [],
     \   '_fulfillments': [],
     \   '_rejections': [],
@@ -90,7 +91,7 @@ function! s:_publish(promise, ...) abort
   endif
 
   if empty(a:promise._children)
-    if settled == s:REJECTED
+    if settled == s:REJECTED && !a:promise._has_floating_child
       call s:_on_unhandled_rejection(a:promise._result)
     endif
     return
@@ -285,8 +286,10 @@ function! s:_promise_then(...) dict abort
   let l:Res = a:0 > 0 ? a:1 : v:null
   let l:Rej = a:0 > 1 ? a:2 : v:null
   if state == s:FULFILLED
+    let parent._has_floating_child = v:true
     call s:Later.call(funcref('s:_invoke_callback', [state, child, Res, parent._result]))
   elseif state == s:REJECTED
+    let parent._has_floating_child = v:true
     call s:Later.call(funcref('s:_invoke_callback', [state, child, Rej, parent._result]))
   else
     call s:_subscribe(parent, child, Res, Rej)
--- a/autoload/vital/__vital__/Async/Promise.vim
+++ b/autoload/vital/__vital__/Async/Promise.vim
@@ -90,6 +90,9 @@ function! s:_publish(promise, ...) abort
   endif
 
   if empty(a:promise._children)
+    if settled == s:REJECTED
+      call s:_on_unhandled_rejection(a:promise._result)
+    endif
     return
   endif
 
@@ -268,6 +271,11 @@ function! s:wait(promise, ...) abort
   else
     return [v:null, a:promise._result]
   endif
+endfunction
+
+let s:_on_unhandled_rejection = s:NOOP
+function! s:on_unhandled_rejection(on_unhandled_rejection) abort
+  let s:_on_unhandled_rejection = a:on_unhandled_rejection
 endfunction
 
 function! s:_promise_then(...) dict abort
--- a/doc/vital/Async/Promise.txt
+++ b/doc/vital/Async/Promise.txt
@@ -377,6 +377,13 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
 	call Promise.wait(p, 1000)
 	" Is equivalent to call Promise.wait(p, {'timeout': 1000})
 >
+on_unhandled_rejection({callback})	*Vital.Async.Promise.on_unhandled_rejection*
+
+ 	Set callback to catch all unhandled rejected promise's result.
+
+	Note:
+	This callback will called even if you using |Vital.Async.Promise.wait()|.
+
 
 is_promise({value})			*Vital.Async.Promise.is_promise()*
 
--- a/doc/vital/Async/Promise.txt
+++ b/doc/vital/Async/Promise.txt
@@ -380,6 +380,7 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
 on_unhandled_rejection({callback})	*Vital.Async.Promise.on_unhandled_rejection*
 
  	Set callback to catch all unhandled rejected promise's result.
+	If {callback} throws error, |Async.Promise| does not handle it.
 
 	The {callback} is |Funcref|, it's argument can be unhandled thrown error or unhandled rejected value.
 
--- a/doc/vital/Async/Promise.txt
+++ b/doc/vital/Async/Promise.txt
@@ -381,9 +381,15 @@ on_unhandled_rejection({callback})	*Vital.Async.Promise.on_unhandled_rejection*
 
  	Set callback to catch all unhandled rejected promise's result.
 
+	The {callback} is |Funcref|, it's argument can be unhandled thrown error or unhandled rejected value.
+
 	Note:
 	This callback will called even if you using |Vital.Async.Promise.wait()|.
+	If you want to clear callback, you can use following codes.
 
+>
+	call Promise.on_unhandled_rejection(Promise.noop)
+<
 
 is_promise({value})			*Vital.Async.Promise.is_promise()*
 
--- a/test/Async/Promise.vimspec
+++ b/test/Async/Promise.vimspec
@@ -804,6 +804,56 @@ Describe Async.Promise
       endif
     End
   End
+
+  Describe .on_unhandled_rejection
+    It should call when promsie does not catch
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> execute('throw "error"') })
+      call s:wait_has_key(l, 'result')
+      Assert HasKey(result, 'exception')
+      Assert HasKey(result, 'throwpoint')
+    End
+
+    It should call when promise does not catch with finally
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> execute('throw "error"') }).finally({ -> {} })
+      call s:wait_has_key(l, 'result')
+      Assert HasKey(result, 'exception')
+      Assert HasKey(result, 'throwpoint')
+    End
+
+    It should call when promise does not catch with children
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> execute('throw "error"') }).then({ -> {} })
+      call s:wait_has_key(l, 'result')
+      Assert HasKey(result, 'exception')
+      Assert HasKey(result, 'throwpoint')
+    End
+
+    It should call when promise does not catch with wait
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> execute('throw "error"') }).then({ -> {} })
+      let [_, error] = P.wait(p)
+      Assert Equals(error, result)
+    End
+
+    It should not call when promise catched
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> execute('throw "error"') }).catch({ -> {} })
+      call P.wait(Wait(100))
+      Assert KeyNotExists(l, 'result')
+    End
+  End
 End
 
 " vim:et ts=2 sts=2 sw=2 tw=0:
--- a/test/Async/Promise.vimspec
+++ b/test/Async/Promise.vimspec
@@ -864,6 +864,15 @@ Describe Async.Promise
       let l = l:
       call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
 
+      let p = P.reject({ 'error': 'error' }).catch({ -> {} })
+      call P.wait(Wait(100))
+      Assert KeyNotExists(l, 'result')
+    End
+
+    It should not call when promise catched
+      let l = l:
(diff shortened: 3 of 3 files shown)


doc/vital/Async/Promise.txt
```diff
@@ -377,6 +377,13 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
 	call Promise.wait(p, 1000)
 	" Is equivalent to call Promise.wait(p, {'timeout': 1000})
 >
+on_unhandled_rejection({callback})	*Vital.Async.Promise.on_unhandled_rejection*
+
+ 	Set callback to catch all unhandled rejected promise's result.
```
rhysd: Would you clarify the function signature of `{callback}`?

test/Async/Promise.vimspec
```diff
@@ -804,6 +804,56 @@ Describe Async.Promise
...
+    End
+
+    It should not call when promise catched
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> execute('throw "error"') }).catch({ -> {} })
+      call P.wait(Wait(100))
+      Assert KeyNotExists(l, 'result')
+    End
```
rhysd: Please add a test case where `on_unhandled_rejection` is set but the callback is not called since the promise was handled.

doc/vital/Async/Promise.txt
```diff
@@ -377,6 +377,13 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
...
+
+	Note:
+	This callback will called even if you using |Vital.Async.Promise.wait()|.
```
rhysd: Would you also add note how to clear the callback.

test/Async/Promise.vimspec
```diff
@@ -804,6 +804,56 @@ Describe Async.Promise
       endif
     End
   End
+
+  Describe .on_unhandled_rejection
```
rhysd: Please add `After` hook which clears a callback set at `on_unhandled_rejection`. If we add another tests after this `Describe` block, they will be affected by callbacks set in test cases in this `Describe` block

autoload/vital/__vital__/Async/Promise.vim
```diff
@@ -90,6 +90,9 @@ function! s:_publish(promise, ...) abort
   endif
 
   if empty(a:promise._children)
+    if settled == s:REJECTED
+      call s:_on_unhandled_rejection(a:promise._result)
```
rhysd: Question: Should we handle an exception when this call throws?

test/Async/Promise.vimspec
```diff
@@ -804,6 +804,56 @@ Describe Async.Promise
...
+    It should call when promsie does not catch
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> execute('throw "error"') })
+      call s:wait_has_key(l, 'result')
+      Assert HasKey(result, 'exception')
+      Assert HasKey(result, 'throwpoint')
```
rhysd: Would you add test case where promise rejected without `throw`? Current all test cases make rejected promise values by `throw`.

```vim
call P.on_unhandled_rejection(...)

P.reject('hi')

" Verify the callback is called with correct argument
```

test/Async/Promise.vimspec
```diff
@@ -804,6 +804,56 @@ Describe Async.Promise
       endif
     End
   End
+
+  Describe .on_unhandled_rejection
+    It should call when promsie does not catch
```
rhysd: nit: typo

autoload/vital/__vital__/Async/Promise.vim
hrsh7th: Thanks for your all reviews.

Hm... In my opinion, If we catch the error in here, we can only echo it.

```viml
      try
        call s:_on_unhandled_rejection(a:promise._result)
      catch /.*/
        echomsg string({ 'exception': v:exception, 'throwpoint': v:throwpoint })
      endtry
```

By default, `s:_on_handled_rejection` is `s:noop` that does not throw error.
If user set custom `s:_on_unhandled_rejection`, it has the possibility to throw an error.

In this case, the user wants to know thrown error, I think.

Should I add echomsg in here?

test/Async/Promise.vimspec
hrsh7th: Thanks for your review.
I've added test.

test/Async/Promise.vimspec
hrsh7th: Thanks for your review.
I've added test.

test/Async/Promise.vimspec
hrsh7th: It is my bad.
I've added after hook.

doc/vital/Async/Promise.txt
hrsh7th: I've added docs to clarify how to clear callback.

doc/vital/Async/Promise.txt
hrsh7th: I've added callback signature docs.

test/Async/Promise.vimspec
```diff
@@ -804,6 +804,80 @@ Describe Async.Promise
...
+      Assert KeyNotExists(l, 'result')
+    End
+
+    It should not call when promise does not throw error
+      let l = l:
+      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })
+
+      let p = P.resolve().then({ -> { 'success': 'success' } })
+      call P.wait(Wait(100))
+      Assert KeyNotExists(l, 'result')
```
rhysd: Thank you for adding this test case. Would you add the below check to here also? It ensures that handled rejection does not trigger the callback.

```vim
let p = P.reject('hi').catch({ -> 'handled' })
call P.wait(Wait(100))
Assert KeyNotExists(l, 'result')
```

autoload/vital/__vital__/Async/Promise.vim
rhysd: Thank you for the consideration. Though I don't have strong opinion here, but we can go with current implementation. Would you add note that user must handle exceptions in the callback at `on_unhandled_rejection` section of documentation? I feel it's enough.

test/Async/Promise.vimspec
hrsh7th: Thanks for your review.

I've added it case on my local environment and then it case is failed...
I investigate the reason... it take a time maybe.

hrsh7th: It seems `P.reject` will call `s:_invoke_callback` immediately.
So `P.reject("hi")` calls `s:_on_unhandled_rejection` before it's catch.

hrsh7th: Hm... I think we can't detect `unhandled rejection` perfectly.

```viml
let P = P.reject("")
call timer_start(1000000, { -> P.catch({ -> {} }) })`
```

The `p` was caught in timer. But it looks like unhandled rejection in a synchronous context.

Node.js notice it when process exited... 
Should we use `VimLeavePre`? 😂

rhysd: > Hm... I think we can't detect unhandled rejection perfectly.

That's intended correct behavior. Let's say we have following JavaScript code:

```js
const p = Promise.reject();

setTimeout(() => p.catch(() => {}), 1000);
```

Running the script also complains as follows:

```
(node:69753) UnhandledPromiseRejectionWarning: undefined
(node:69753) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

rhysd: > It seems P.reject will call s:_invoke_callback immediately.

When calling `P.reject()`, this function is called:

https://github.com/vim-jp/vital.vim/blob/637cae33a3efc5e838f1b736cc436bf13a77733e/autoload/vital/__vital__/Async/Promise.vim#L239-L243

it calls `s:_reject`

https://github.com/vim-jp/vital.vim/blob/637cae33a3efc5e838f1b736cc436bf13a77733e/autoload/vital/__vital__/Async/Promise.vim#L160-L167

I thought this line delays calling `s:_publish` so that succeeding `catch` at `P.reject().catch({-> ''})` works correctly.

```vim
call s:Later.call(funcref('s:_publish', [a:promise]))
```

Does the `s:Later.call` calls `s:_publish` immediately (before calling following `.catch()`)? If so, I think it is a bug. What do you think @lambdalisue?

lambdalisue: > Does the s:Later.call calls s:_publish immediately (before calling following .catch())? If so, I think it is a bug. What do you think @lambdalis

No. It just added to an internal tasks and a task will be called on next tick.
https://github.com/vim-jp/vital.vim/blob/master/autoload/vital/__vital__/Async/Later.vim#L6-L11

lambdalisue:
```vim
let s:Promise = vital#vital#import('Async.Promise')

function! s:test() abort
  echomsg 'aaa'
  let p = s:Promise.reject('oops')
  call p.catch({-> execute('echomsg "ccc"', '') })
  echomsg 'bbb'
endfunction

call s:test()
```

Above produced

```
aaa
bbb
ccc
```

hrsh7th: Thank you for the investigation.
Apparently I have some wrong understanding... I'll more deep investigate it.

rhysd: Thanks. I think I misuderstood something then.

hrsh7th: Sorry. This sample test case was passed. 
The original problem has another reason maybe...

```viml
  It should run reject later
    call add(g:_promise_sequence, 'start')
    call P.reject('hi')
    call add(g:_promise_sequence, 'end')

    call P.wait(Wait(100))

    Assert Equals(g:_promise_sequence, ['start', 'end', 's:_publish']) " put adhoc code in `s:_publish`
  End
```

hrsh7th: FYI: This test case was failed as same as `P.reject(...).catch()` case.
```viml
    It should not call when thrown error catched
      let l = l:
      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })

      let p = P.new({ resolve, reject -> execute('throw "error"') }).catch({ -> {} })
      call P.wait(Wait(100))
      Assert KeyNotExists(l, 'result')
    End
```

rhysd: Thank you for the investigation. It's very helpful. How about following?

```vim
call P.on_unhandled_rejection({ -> add(g:_promise_sequence, 'unhandled') })
call add(g:_promise_sequence, 'start')
call P.reject('hi').catch({ -> add(g:_promise_sequence, 'catch')})
call add(g:_promise_sequence, 'end')

call P.wait(Wait(100))
call add(g:_promise_sequence, 'finish')
```

hrsh7th: The provided test has passed.

```
    It should not call when rejected promise catched aiueo
      call P.on_unhandled_rejection({ -> add(g:_promise_sequence, 'unhandled') })
      call add(g:_promise_sequence, 'start')
      call P.reject('hi').catch({ -> add(g:_promise_sequence, 'catch')})
      call add(g:_promise_sequence, 'end')

      call P.wait(Wait(100))
      call add(g:_promise_sequence, 'finish')

      Assert Equals(g:_promise_sequence, ['start', 'end', 's:_publish', 'unhandled', 'catch', 'finish'])
    End
```

Hmm...

1. `s:_publish` called once.
2. At this time, `children` is empty (So s:_unhandled_rejection called)
3. and catch was called.

I seem to have overlooked something...

hrsh7th: I've understood this problem maybe.

1. `let p = P.reject('hi')` ... `ParentPromise`
- immediately change `p._state` to `s:REJECTED`.
- reserve call `s:_publish`

2. `p.catch({ -> ... })` ... `ChildPromise`
- immediately call `s:_promise_then`
- reserve call `s:_invoke_callback` (because ParentPromise's state is not pending)'
- don't call `s:_subscribe` so ParentPromise has not child.

hrsh7th: Sorry for my poor english.
If my english can't understand, please tell me it.
If so, I will write japanese.

lambdalisue: もう日本語で良い気がします。

そもそもの問題はどういったものなのでしょう？再現コードかけますか？

hrsh7th: 問題を単純化しますと、下記のような状態になっていることが、テストケースの追加でわかったという感じです。それでなぜだろうということで、色々調査をしていました。

```
    It 失敗するケース
      let l = l:
      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })

      let p = P.reject({ 'error': 'error' }).catch({ -> {} })
      call P.wait(Wait(100))
      Assert KeyNotExists(l, 'result')
    End

    It 成功するケース
      let l = l:
      call P.on_unhandled_rejection({ result -> extend(l, { 'result': result }) })

      let p = P.resolve().then({ -> execute('throw "error"') }).catch({ -> {} })
      call P.wait(Wait(100))
      Assert KeyNotExists(l, 'result')
    End
```

hrsh7th: 一度 resolve を挟むか、直接 rejected な promise を経由するかで挙動が変わっている用に見えるというものです。

原因はなんとなく掴めたのかなと思っていて

1. `let p = P.reject('hi')` を実行する
    1. `s:reject` は内部的に `s:_reject` を即座に呼びます。
    2. `s:_reject` は `state を s:REJECTED に即座に変更` して、`s:_publish` の呼び出しを予約します。

2. その後 `p.catch({ -> ...})` が呼ばれます。
    1. 親プロミスは同期的に `s:REJECTED` になっているので、`s:_subscribe` を実行せずに、`s:_invoke_callback` の呼び出しを予約します。

ということで、`s:_subscribe` を `.catch({ -> ... })` で呼び出していないので、親プロミスには子供がいない。という扱いになり `on_unhandled_rejection` が意図せず呼び出されてしまう。

ということがわかった感じです。

hrsh7th: まだ深堀りできてないですが、いろんな Promise の実装を見てみたところ、`state を s:REJECTED に即座に変更` は一般的な実装のように思えました。

ので、on_unhandled_rejection 側の考慮が足りてない。ということになりそうです。
どう治すべきなのか、検討をしてからになりそうなのでこの PR は進みが悪くなりそうです。。。

hrsh7th: おそらくこの手法は bluebird 方式とほぼ同じ効力な気がします。
https://github.com/petkaantonov/bluebird/blob/master/src/promise.js#L285

rhysd: 調査＆修正ありがとうございます。その修正方法で完璧だと思います。

hrsh7th:
無事マージまで持って行けてよかったです。

@rhysd @lambdalisue レビュー・調査、諸々本当にありがとうございました！

PR #722, opened 2020-02-14
System.Filepath: Introduce expand_home() function

bb010g:
The `:p` absolute path filename modifier supports leading «"~/" (and "~user/" for Unix)» for home directories. As a home-relative path like `~` does not resolve properly with filereadable() or isdirectory() (unless there's a subdirectory named `~` in the current working directory! 😅), we need to expand it ourselves.

Note that the docs say just `~/` is changed, but `~` works just as well.

---
This change is [Reviewable](https://reviewable.io/reviews/vim-jp/vital.vim/722)

### DIFF
--- a/autoload/vital/__vital__/System/Filepath.vim
+++ b/autoload/vital/__vital__/System/Filepath.vim
@@ -197,12 +197,23 @@ function! s:is_case_tolerant() abort
 endfunction
 
 
+function! s:expand_home(path) abort
+  if a:path[:0] !=# '~'
+    return a:path
+  endif
+  let post_home_idx = match(a:path, s:path_sep_pattern)
+  return post_home_idx is# -1
+        \ ? s:remove_last_separator(expand(a:path))
+        \ : s:remove_last_separator(expand(a:path[0 : post_home_idx - 1]))
+        \   . a:path[post_home_idx :]
+endfunction
+
 function! s:abspath(path) abort
   if s:is_absolute(a:path)
     return a:path
   endif
-  " Note:
-  "   the behavior of ':p' for non existing file path/directory is not defined
+  " NOTE: The behavior of ':p' for a non existing file/directory path
+  " is not defined.
   return (filereadable(a:path) || isdirectory(a:path))
         \ ? fnamemodify(a:path, ':p')
         \ : s:join(fnamemodify(getcwd(), ':p'), a:path)
--- a/doc/vital/System/Filepath.txt
+++ b/doc/vital/System/Filepath.txt
@@ -97,6 +97,15 @@ is_case_tolerant()		*Vital.System.Filepath.is_case_tolerant()*
 	Return non-zero if filesystem ignores alphabetic case of a filename,
 	zero otherwise.
 
+expand_home({path})		*Vital.System.Filepath.expand_home()*
+	Return a home-expanded path of {path}.  The home directory expansion
+	will not have a trailing directory separator.
+
+	Like |expand()|, but only expanding a leading "~" home directory
+	sequence, or like |fnamemodify()| with the ":p" modifier, but without
+	full path expansion or unpredictable results if a file name doesn't
+	exist and doesn't have an absolute path.  See |filename-modifiers|.
+
 abspath({path})			*Vital.System.Filepath.abspath()*
 	Return an absolute path of {path}.
 	If the {path} is already an absolute path, it returns the {path}.


autoload/vital/__vital__/System/Filepath.vim
```diff
@@ -271,4 +274,4 @@ endfunction
 let &cpo = s:save_cpo
 unlet s:save_cpo
 
-" vim:set et ts=2 sts=2 sw=2 tw=0:
```
tsuyoshicho: Do not need change modeline, revert, please.

bb010g: Sorry, changed that hoping that this file's less normal modeline was what made GitHub's integrated editor not syntax highlight. But no change in detection with a modeline that's more normalized, and it looks like the editor doesn't syntax highlight at all, actually.

autoload/vital/__vital__/System/Filepath.vim
```diff
@@ -198,14 +198,17 @@ endfunction
 
 
 function! s:abspath(path) abort
-  if s:is_absolute(a:path)
-    return a:path
+  let path = substitute(a:path,
+  \                     '^\(\~.\{-}\)\%($\|' . s:path_sep_pattern . '\)',
+  \                     {m -> fnamemodify(m[1], ':p')}, '')
```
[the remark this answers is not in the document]
bb010g: `expand()` has the issue of also handling environment variables & its special keywords. Would this functionality be alright in an `s:expand_home()` function?

I think some sort of regex would want to be kept, however, as you wouldn't want to pass too much to `expand()`. `'^\(\~.\{-}\)\(' . s:path_sep_pattern . '\)\?'` would work (match up to the first separator run found, process first group, retain second group as-is), but it's still probably not great? `'^\~' . s:not_path_sep_pattern` (match & process up to first separator character) would work if we introduced a negated companion to `s:path_sep_pattern`.

Avoiding `substitute()`, the following function would also work:
```vim
function! s:expand_home(path) abort
  let idx = match(a:path, s:path_sep_pattern)
  let path = s:remove_last_separator(expand(a:path[0 : idx]))
  if idx > 0
    let path += path[idx + 1 :]
  endif
  return path
endfunction
```

autoload/vital/__vital__/System/Filepath.vim
```diff
@@ -198,14 +198,17 @@ endfunction
...
+  if s:is_absolute(path)
+    return path
```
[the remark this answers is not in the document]
bb010g: This is fair, especially without a flag to disable home resolution, since other functions can use paths like `'~'` without resolving them to home.

autoload/vital/__vital__/System/Filepath.vim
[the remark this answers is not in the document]
lambdalisue: I thought you've need a "~ expanded path" for `filereadable()` and `isdirectory()` thus what I meant was like (I noticed that the code below does not work when `a:path` points missing file thought.)

```vim
function! s:abspath(path) abort
  if a:path ==# '' || s:is_absolute(a:path)
    return a:path
  endif
  let expanded = expand(a:path)
  return filereadable(expanded) || isdirectory(expanded)
        \ ? fnamemodify(a:path, ':p')
        \ : s:join(fnamemodify(getcwd(), ':p'), a:path)
endfunction
```

Additionally, if you want to expand just `~`, you don't need regex to replace `~` in `a:path` like

```vim
function! s:expand_home(path) abort
  return a:path ==# '~' || a:path[:1] ==# '~' . s:separator()
        ? fnamemodify(a:path, ':p')
        : a:path
endfunction
```

I really appreciate your contribution but I don't think handling `~` is not a business of `abspath()` in my opinion. I'm sorry about this. Please convince me if my argument is off the point.

bb010g:
Okay, pivoted this patch into creating a new `expand_home({path})` function. Commit message:

Because sometimes you want to expand the path `"~vim-jp/$EDITOR"` to `"/home/vim-jp/$EDITOR"`, but that file name doesn't exist, making `fnamemodify(_, ':p')` unpredictable, and the environment variable syntax in the path makes `expand(_)` unwieldy.

This should be useful while not adding complexity to or forcing unwanted expansions from `abspath()`.

Also, English documentation has been provided.

autoload/vital/__vital__/System/Filepath.vim
```diff
@@ -197,12 +197,23 @@ function! s:is_case_tolerant() abort
 endfunction
 
 
+function! s:expand_home(path) abort
+  if a:path[0] !=# '~'
```
lambdalisue: *should*

Use `[:0]` to avoid exceptions when `path` is an empty string.

```vim
if a:path[:0] !=# '~'
```

autoload/vital/__vital__/System/Filepath.vim
```diff
@@ -197,12 +197,23 @@ function! s:is_case_tolerant() abort
...
+    return a:path
+  endif
+  let post_home_idx = match(a:path, s:path_sep_pattern)
+  return post_home_idx == -1
+        \ ? s:remove_last_separator(expand(a:path))
+        \ : s:remove_last_separator(expand(a:path[0 : post_home_idx - 1]))
+        \   . a:path[post_home_idx :]
+endfunction
```
lambdalisue: may

If you want to expand a leading `~/` or `~\`, I think simple replacement like below is enough and better

```vim
let sep = s:separator()
if a:path[:1] ==# '~' . sep
  return s:join(expand('~'), a:path[2:])
elseif a:path[:0] ==# '~'
  return s:join(expand('~'), a:path[1:])
endif
return a:path
```

bb010g: I like this, but Vim still supports `~username` syntax, and I think we'd want to support it here.

I.e., even if I'm logged in as the user `bb010g` with a home directory of `/home/dusk`, and you're the user `lambdalisue` with a home directory of `/home/alisue` on the same machine, the path `~lambdalisue/.config/nvim/init.vim` will expand for me to `/home/alisue/.config/nvim/init.vim`. This function would leave the former unexpanded, and only let me expand as the user `bb010g` the path `~/.config/nvim/init.vim` to `/home/dusk/.config/nvim/init.vim`.

lambdalisue: I didn't know that 👍 Well then, how about

```vim
return post_home_idx is# -1
      \ ? a:path
      \ : s:join(expand(a:path[: post_home_idx - 1]), a:path[post_home_idex:])
```

- `is#` is a bit better to compare numbers
- I think we should not touch `a:path` because the function is to expand ONLY leading `~` right?
- We can use `s:join()` instead of combination of `s:remove_last_separator()` and `.`

bb010g: >     * I think we should not touch `a:path` because the function is to expand ONLY leading `~` right?

If there's no leading `~`, we've already returned the path unmodified. So no match means no path separators in the path, and we've got a path like `~` or `~vim-jp` instead of `~/` or `~vim-jp/`. We still need to home-expand these paths.

>     * We can use `s:join()` instead of combination of `s:remove_last_separator()` and `.`

We could, but `s:join()` seems overly heavy here?

I originally didn't think we'd need `s:remove_last_separator()`, but testing on available Linux & Windows shells resulted in `/home/bb010g` on Linux (no trailing sep) & `C:\Users\bb010g\` on Windows (trailing sep). Focusing on just expanding `~[…]`, pre–any separator, like a hypothetical `home_path([{user}])` function, we can normalize the result to no trailing separator. So, a path of `~[{user}]` becomes `/home/{user}` or `C:\User\{user}`, and we're done. In the case that the user *did* want a trailing separator, we have the advantage of the whole path string, and can just use whatever they specify. So `~foo/` becomes `/home/foo/` (parts `/home/foo` & `/`) and `~\` becomes `C:\Users\vim-jp\` (parts `C:\Users\vim-jp` & `\`).

If we were to pull in `s:join()` for the second case (the only case in which it makes sense), we'd now be:
* Calling `s:separator()` (which invokes a `|::p|` expansion on the current directory)
* Allocating a new empty string
* Looping over all one arguments to:
  * Type check the argument, a known string
  * Create a new string from concatenating it & the empty string
* Calling substitute over a string slice to match every `s:path_sep_pattern` and replace them with the earlier `s:separator()` result

All of this seems pretty useless when we can do less work, reasonably, and prefer the user's input. We gain a purer expansion too that doesn't muck with directory separators, unless the user wants to themselves. And, if the output of this already being joined with other directories later, we'd just be duplicating work. For the case of environment variables like `XDG_CACHE_HOME` that can contain home directory sequences that Vim already handles properly (but your user code may not), I'd argue not mucking with path separators is beneficial, as it gets you closer to the Vim behavior that you're trying to emulate.

lambdalisue: > If there's no leading ~, we've already returned the path unmodified. So no match means no path separators in the path, and we've got a path like ~ or ~vim-jp instead of ~/ or ~vim-jp/. We still need to home-expand these paths.

Oops. Sorry I misunderstood here. Got it 👍 

>  In the case that the user did want a trailing separator, we have the advantage of the whole path string, and can just use whatever they specify.

Good point. Now I think I understand why you've use `s:remove_last_separator()` for `expand(a:path)` part as well. So you've tried to make consistent results like below right?

```vim
echo s:expand_path("foo/bar")
" foo/bar

echo s:expand_path("~")
" /home/lambdalisue

echo s:expand_path("~vim-jp")
" /home/vim-jp
" C:\Users\vim-jp

echo s:expand_path("~vim-jp/")
" /home/vim-jp/
echo s:expand_path("~vim-jp/foo")
" /home/vim-jp/foo
echo s:expand_path("~vim-jp/foo/")
" /home/vim-jp/foo/

echo s:expand_path("~vim-jp\\")
" C:\Users\vim-jp\
echo s:expand_path("~vim-jp\\foo")
" C:\Users\vim-jp\foo
echo s:expand_path("~vim-jp\\foo\\")
" C:\Users\vim-jp\foo\
```

I just want to make sure if I understand it correctly. It's much complicated than I thought.

lambdalisue:
@bb010g Thanks for you contribution 🎉

autoload/vital/__vital__/System/Filepath.vim
bb010g: Yep! And by operating in this manner, we save on computation in case the user wants to call this function a lot.

PR #743, opened 2020-05-21
Web.JSON: fix char encodes

Milly:
This is part of #695, only bug fixes.

* Encode all control chars.
  * Prev: only `\r`, `\n` and `\t` are converted.
  * Fixed: all of `\x01` to `\x1f` are converted.
* Decode UTF-16 surrogate pairs.
  * Prev: can not decoded.
  * Fixed: can decoded.


Milly:
It is unrelated failures:
- Deprecated.Text.Sexp parse
- Async.Later .call() calls tasks in FIFO manner (nest)

PR #744, opened 2020-06-08
vital: fix import error handling

Milly:
Vital shows stacktrace, when error throws in `module._vital_loaded()`.
But the stacktrace formatter function has bugs.

* If anonymous or lambda function exists in stack, an error occurs in the formatter.
* Function attributes are not detected.

... and minor fixes.


Milly:
rebased master

PR #751, opened 2020-08-02
autoloadablize.vim: SID hack use expand('<SID>')

tsuyoshicho:
Change to expand() after 8.2.1347.

Use if need.


lambdalisue (commented):
It seems supporting `expand('<SNR>')` actually introducing additional complexity. I think it would be better to stick to the current implementation.

tsuyoshicho:
Okay, This PR is closed, and keep PR branch store information.

PR #752, opened 2020-08-04
Async.Promise: Add "chain" function

lambdalisue:
Vim does not have async/await syntax support thus it's a bit tough work
to write similar function as below (js)

```js
    for (const url of urls) {
        const r = await request(url);
        console.log(r)
    }
```

With this commit, equivalent code of above can be written as

```vim
    let fs = map(urls, { url -> { -> request(url).then({ r -> execute('echo r', '') }) })
    call s:Promise.chain(fs)
```

**Question:** Is there any way to avoid stack-overflow or should I comment on the doc?

### DIFF
--- a/doc/vital/Async/Promise.txt
+++ b/doc/vital/Async/Promise.txt
@@ -379,8 +379,8 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
 >
 chain({promise_factories})		*Vital.Async.Promise.chain()*
 	Chain promises produced by {promise_factories} (|List| of |Function|)
-	and return a result |List| which contains result of each produced
-	promises.
+	and return a promise which resolves to a result |List| which contains 
+	result of each produced	promises.
 	It is like an asynchronous sequential call. It rejects when one of
 	function in {promise_factories} has failed or produced promises
 	rejects. Note that it stops producing promises by functions after
--- a/doc/vital/Async/Promise.txt
+++ b/doc/vital/Async/Promise.txt
@@ -377,6 +377,42 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
 	call Promise.wait(p, 1000)
 	" Is equivalent to call Promise.wait(p, {'timeout': 1000})
 >
+chain({promise_factories})		*Vital.Async.Promise.chain()*
+	Chain promises produced by {promise_factories} (|List| of |Function|)
+	and return a result |List| which contains result of each produced
+	promises.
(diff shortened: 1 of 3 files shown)


doc/vital/Async/Promise.txt
```diff
@@ -377,6 +377,40 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
 	call Promise.wait(p, 1000)
 	" Is equivalent to call Promise.wait(p, {'timeout': 1000})
 >
+chain({promise_factories})		*Vital.Async.Promise.chain()*
+	Chain promises produced by {promise_factories} (|List| of |Function|). 
+	It is like an asynchronous sequential call. It rejects when one of
+	function in {promise_factories} has failed or produced promises
+	rejects. Note that it stop producing promises by functions after
+	rejection.
```
rhysd, line 385: Would you describe what is returned? Documentation only mention the case where it rejects.

doc/vital/Async/Promise.txt
```diff
@@ -377,6 +377,40 @@ wait({promise}[, {options}])		*Vital.Async.Promise.wait()*
 	call Promise.wait(p, 1000)
 	" Is equivalent to call Promise.wait(p, {'timeout': 1000})
 >
+chain({promise_factories})		*Vital.Async.Promise.chain()*
+	Chain promises produced by {promise_factories} (|List| of |Function|). 
+	It is like an asynchronous sequential call. It rejects when one of
+	function in {promise_factories} has failed or produced promises
+	rejects. Note that it stop producing promises by functions after
```
rhysd, line 384: typo: stop -> stops

lambdalisue:
@rhysd Updated. The CI fails but not related to this fix

tyru:
@lambdalisue 
Just wondering, I don't want to block this PR.

What is `async for` ? You mean this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

Did you intend `do requests in parallel, then output the results in order` ?

tyru:
Sorry, doc already mention that. Please ignore.

lambdalisue:
My bad. As you said `for await` is correct in JS. The `async for` is Python sorry.

lambdalisue:
I've squashed commits and added https://github.com/vim-jp/vital.vim/pull/752/commits/5d00a5c9536ca6c3403190cac3f201282c8d7b96 to test that the number of factories can be beyond the maxfuncdepth.

PR #756, opened 2020-09-20
System.Process.Vimproc: fix timeout unit

tsuyoshicho:
## in vimproc help

```
vimproc#system({expr} [, {input} [, {timeout}]])	*vimproc#system()*

(...)

		If you set {timeout}, vimproc will kill the process after
		{timeout} and throw "vimproc: vimproc#system(): Timeout."
		exception.
		The unit is millisecond.
```

timeout unit as millisecond.

## in System.Process.execute() help

```
				*Vital.System.Process.execute()*
execute({args}[, {options}])

(...)

	The following attributes are supported in {options}

(...)

		"timeout"
		A process timeout in seconds (|Number|). A system process
		client does not support this option, mean that vimproc is
		required to be available.
		Default: 0
```

timeout unit as second.

## current implementation.

https://github.com/vim-jp/vital.vim/blob/1b9fab508176efd8bc3fcbc435afd8d7b928ef53/autoload/vital/__vital__/System/Process/Vimproc.vim#L37-L70

call `vimproc#system()` with options.timeout directly.

use second value to millisecond.

This PR fix it(and test value).
But this fixing are breaking change.

After PR test done, `Changes` changing commit add.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,3 +1,7 @@
+8ebc53cf1ed6c53ee793e31611829fa57ae7c06d
+	Modules: System.Process.Vimproc
+	Documents as wrote seconds, but it was actually treated as milliseconds.
+	Fix inner timeout unit, timeout set as seconds, call function as milliseconds at value x1000.
 97e8cf8b230cd69720e130742859b51f5761196c
 	Modules: Web.HTTP
 	Added python3 to HTTP client, and refined its default value.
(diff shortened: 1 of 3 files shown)


tsuyoshicho:
Done.

Test MacVim fail need ignore. (known issue)

lambdalisue (approved):
LGTM

PR #763, opened 2020-11-15
Data.BigNum: test expand

tsuyoshicho:
Test type convert to vimspec.

And Test case expand, fill some case.

## Add more test

- from_num : invalid string, but valid string data.
- from_string : invalid string not number / float.
- add : positive value, positive value.
- sub : positive value, positive value. result positive.
- sub : positive value, positive value. result negative.
- sub : negative value, negative value. result negative.
- mul : positive value, negative value.
- div_mod : basic div/mod, div zero.
- neg : positive value.
- neg : huge negative to positive.


ujihisa (approved):
Hmm, I thought we are converting from .vimspec tests to .vim tests. What's your thought @thinca?

tsuyoshicho:
Thanks review.

The vimspec format is easy to see and I think it will be useful for new feature additions and test expansion.

thinca:
Which one to choose is the maintainer's preference.

tsuyoshicho:
Thanks all.

PR #773, opened 2021-10-02
fix: Web.HTTP: outputFile in Windows like tempfile

tsuyoshicho:
## Detail
Error happen Vim with https://github.com/lambdalisue/vim-protocol in Windows.

Researched to vim-protocol and vital.vim, Web.HTTP with `client outputFile`.
Resolved issue.

## Reason
curl/wget support outputFile. and these clients use tempfile in client implements.

Inner tempfile was successful but external supply were failed.
Differenent both files are using `tr()` that replace from backslash to slash. 

This PR  apply `tr()` to external supply `outputFile` param.

## Test

local vim-protocol with replaced `outputFile` (in HTTP.request param) and updated vital are success.

### DIFF
--- a/autoload/vital/__vital__/Web/HTTP.vim
+++ b/autoload/vital/__vital__/Web/HTTP.vim
@@ -172,7 +172,11 @@ function! s:_make_postfile(data) abort
 endfunction
 
 function! s:_tempname() abort
-  return tr(tempname(), '\', '/')
+  return s:_file_resolve(tempname())
+endfunction
+
+function! s:_file_resolve(file) abort
+  return tr(a:file, '\', '/')
 endfunction
 
 function! s:_postdata(data) abort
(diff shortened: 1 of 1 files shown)


tsuyoshicho:
@lambdalisue san.

WHen this problem will be resolved, then next work for vim-protocol update vital PR.
Please review it.

autoload/vital/__vital__/Web/HTTP.vim
```diff
@@ -172,7 +172,11 @@ function! s:_make_postfile(data) abort
 endfunction
 
 function! s:_tempname() abort
-  return tr(tempname(), '\', '/')
+  return s:_file_resolve(tempname())
+endfunction
+
+function! s:_file_resolve(file) abort
+  return tr(a:file, '\', '/')
```
lambdalisue, line 179: Maybe `fnamemodify` is more common for this usecase while that's in help (`:help filename-modifiers`)

```vim
  return fnamemodify(a:file, 'gs?\\?/?')
```

lambdalisue (approved):
I'm not sure if this is required but if you say it is, I'll say

image

autoload/vital/__vital__/Web/HTTP.vim
tsuyoshicho: Thanks, and try it.
It tuned.

```suggestion
  return fnamemodify(a:file, ':gs?\\?/?')
```

tsuyoshicho:
Review requested.

PR #777, opened 2021-10-13
fix: lint error suppress

tsuyoshicho:
All lint error suppress.

in forked master result
https://github.com/tsuyoshicho/vital.vim/runs/3889309816?check_suite_focus=true

- vimlint : do not use variables, suppressed
- vint : blob literal, workarounded


ujihisa:
:slot_machine:

